PR #23, opened 2015-10-05
Removes the loop wrapper around the library

gdraynz:


lerignoux:
Need also to update all nyukis : 
from nyukiland
app reminder nyuki.

gdraynz:
Nyuki is versionned, no need to modify anything else when we merge something.

PR #27, opened 2015-11-03
Add a swagger doc route

gdraynz:


nyuki/nyuki.py
```diff
@@ -232,3 +233,14 @@ def patch(self, request):
...
             return Response(self._config)
+
+    @resource(endpoint='/swagger', version='v1')
+    class Swagger:
+        def get(self, request):
+            try:
+                with open('swagger.json', 'r') as f:
+                    body = json.loads(f.read())
+            except OSError:
+                return Response(status=404)
```
lerignoux: let's tell that is because swagger.json was not found !

gdraynz: done

PR #36, opened 2015-12-07
Allow empty responses in nyuki

gdraynz:


nyuki/capabilities.py
```diff
@@ -37,20 +39,35 @@ def __hash__(self):
...
+
+    def _valid_body(self):
+        if not self.body:
+            return
+
         if not isinstance(self.body, dict) and not isinstance(self.body, list):
             raise ValueError("Response body should be a dictionary or a list")
+
+        # If body and body is json, set the right headers
+        self.headers = {'Content-Type': 'application/json'}
```
Ketouem: `self.body` could be either a `dict` or `list` and yet not be json serializable.

```python
import json

class Test:
    ATTR = 'VALUE'

t = Test()
json.dumps({'test': t})
```

which would give `TypeError: <__main__.Test object at 0x103c30dd8> is not JSON serializable`

nyuki/capabilities.py
```diff
@@ -37,20 +39,35 @@ def __hash__(self):
...
     ENCODING = 'utf-8'
 
     def __init__(self, body=None, status=200):
-        self.body = body or dict()
+        self.body = body
         self.status = status
+        self.headers = {}
         self._is_valid()
 
     def _is_valid(self):
```
Ketouem: The `_is` prefix implies that this method will return a boolean whereas it raises exception when a validation error occurs. Something like `_validate` may be more appropriate ?

nyuki/capabilities.py
```diff
@@ -37,20 +39,35 @@ def __hash__(self):
...
+        self._valid_body()
+        self._valid_status()
+
+    def _valid_body(self):
```
Ketouem: Same as above, resp. with `_validate_body`.

nyuki/capabilities.py
```diff
@@ -37,20 +39,35 @@ def __hash__(self):
...
+        if not self.body:
+            return
+
         if not isinstance(self.body, dict) and not isinstance(self.body, list):
             raise ValueError("Response body should be a dictionary or a list")
+
+        # If body and body is json, set the right headers
+        self.headers = {'Content-Type': 'application/json'}
+
+    def _valid_status(self):
```
Ketouem: Same as above, resp. with `_validate_status`.

nyuki/capabilities.py
gdraynz: I don't think we should handle everything and should just let the response break if the dev tries to return non-serializable values. `<object> is not JSON serializable` will be a good enough error message.

PR #40, opened 2015-12-21
Service management

gdraynz:
This requires a first review: it will be better to look directly at the end result and not the entire diff.
Services like bus or api are now Subclasses of `Service` which give some guidelines to be used inside Nyuki.
Example to add the bus service to a nyuki :
```python
self.services.add('bus', Bus(self))
self.bus.configure(**self.config['bus'])
self.bus.start()
```
A first step to a more modular nyuki library


examples/sample.py
```diff
@@ -28,17 +27,14 @@ def __init__(self):
...
         }
 
-    def setup(self):
+    async def setup(self):
         log.info("Oh great, I'm connected and ready to do what I want!")
-
-    @on_event(Event.Connected)
-    def _on_start(self):
-        self.subscribe('sender')
+        self.bus.subscribe('sender')
```
lerignoux: what is the default method called ?

nyuki/bus.py
```diff
@@ -138,31 +182,25 @@ def _on_event(self, event):
...
-        """
-        Disconnect from the bus with a default timeout set to 5s.
-        """
-        if not self.client.transport:
-            log.warning('XMPP client is already disconnected')
-            self.client.disconnected.set_result(True)
+        callback = self._callbacks.get(efrom)
+        if callback:
+            if not asyncio.iscoroutinefunction(callback):
+                log.warning('event callback must be a coroutine')
```
lerignoux: we don't know what is not a coroutine, will be a hassle to debug, why not add a debug with the problematic coroutine so that we can find it if needed

requirements.txt
```diff
@@ -1,3 +1,4 @@
 aiohttp>=0.19,<0.20
+asyncio_redis>=0.14,<0.15
 jsonschema>=2.5,<2.6
-slixmpp>=1.1,<1.2
+websockets>=2.7,<2.8
```
lerignoux: needed ?

lerignoux:
need a rebase

lerignoux:
for such big changes it might have been better to move one "service"
and the do the following ones in different PR, that would be easier to reveiw

examples/sample.py
gdraynz: This is not used neither in master nor this branch, will fix

nyuki/capabilities.py
```diff
@@ -22,10 +23,23 @@ def decorated(cls):
     return decorated
 
 
+def websocket(type):
```
lerignoux: to remove

nyuki/bus.py
```diff
@@ -50,31 +53,81 @@ class Bus(object):
...
+        self._nyuki.register_schema(self.CONF_SCHEMA)
+        self._loop = self._nyuki.loop or asyncio.get_event_loop()
+
+        self.client = None
+        self._connected = asyncio.Event()
+        self._disconnected = asyncio.Event()
+
+        self._callbacks = dict()
+        self._topic = None
+        self._mucs = None
```
lerignoux: shouldn't it be an empty list

nyuki/commands.py
```diff
@@ -18,7 +18,7 @@ def _build_args():
     parser.add_argument('-c', '--config',
                         help='config file name', required=False)
     parser.add_argument('-j', '--jid',
-                        help='xmpp jid: <user>[@<host>]', required=False)
+                        help='xmpp jid', required=False)
```
lerignoux: to put back

nyuki/logs.py
```diff
@@ -29,9 +29,6 @@
         },
         "slixmpp": {
             "level": "WARNING"
-        },
-        "aiohttp.web": {
```
lerignoux: check

nyuki/nyuki.py
```diff
@@ -158,33 +130,31 @@ def _stop_loop(self):
...
 
-        def timed_out():
-            log.warning('Could not stop nyuki after '
-                        '{} seconds, killing'.format(timeout))
-            self._stop_loop()
-
-        self.loop.call_later(timeout, timed_out)
-        self._exposer.shutdown()
-        self._bus.disconnect()
+    async def report_error(self, code, message):
```
lerignoux: to remove

tests/bus_test.py
```diff
@@ -36,144 +33,41 @@ def test_001b_init(self):
...
         future.set_exception(IqTimeout(None))
-        m = Mock()
+        m = MagicMock()
         m.add_done_callback = lambda f: f(future)
         send_mock.return_value = m
         self.bus._on_register(None)
-        assert_in(Event.ConnectionError, self.events)
-
-
-class TestBusRequest(AsyncTestCase):
```
lerignoux: to restore

lerignoux:
missing Services and Service Tests

PR #41, opened 2015-12-23
Set stdout codec as utf-8

gdraynz:


nyuki/nyuki.py
```diff
@@ -50,6 +53,13 @@ class Nyuki(metaclass=MetaHandler):
     }
 
     def __init__(self, **kwargs):
+        # Set stdout as utf-8 codec
+        try:
+            sys.stdout = codecs.getwriter('utf-8')(sys.stdout.detach())
+        except Exception:
+            # Nosetests seems to alter stdout, breaking detach()
+            log.warning('Could not change stdout codec')
```
lerignoux: no we need the exception at least !
log.exception(exc)

PR #44, opened 2016-01-06
Ensuring the prosody certificate is recognized

lerignoux:


nyuki/bus.py
```diff
@@ -29,6 +29,9 @@ def __init__(self, jid, password, host=None, port=5222):
         # Disable IPv6 until we really need it
         self.use_ipv6 = False
 
+        # Check the server certificate:
+        self.ca_certs = "./ssl/prosody.crt"
```
lhavel: I think that should be configurable for more flexibility. As agreed I'll create a new issue for that.

PR #49, opened 2016-01-08
Make use of a real aiohttp Response object on nyuki API

gdraynz:
Allow us to send back any `content-type` and such

### DIFF
--- a/nyuki/capabilities.py
+++ b/nyuki/capabilities.py
@@ -1,6 +1,7 @@
+from aiohttp import web
+import asyncio
 import json
 import logging
-import asyncio
 
 from nyuki.api import Api
 from nyuki.services import Service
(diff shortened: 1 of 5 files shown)


nyuki/capabilities.py
```diff
@@ -41,48 +42,31 @@ def __hash__(self):
...
-        if self.body:
-            payload = json.dumps(self.body)
-            return bytes(payload, self.ENCODING)
-        return None
+    def __init__(self, body=None, **kwargs):
+
+        # Check json
+        if isinstance(body, dict) or isinstance(body, list):
+            log.debug('converting dict/list response to bytes')
+            body = json.dumps(body).encode()
```
lerignoux: restore ENCODING

nyuki/capabilities.py
```diff
@@ -41,48 +42,31 @@ def __hash__(self):
...
+            if not self._get_content_type(kwargs):
+                kwargs['content_type'] = 'application/json'
+        # Check body
+        elif body is not None:
+            log.debug('converting string response to bytes')
+            body = str(body).encode()
```
lerignoux: restore ENCODING

PR #59, opened 2016-02-01
Websockets

gdraynz:


nyuki/web.py
```diff
@@ -0,0 +1,187 @@
+import asyncio
```
lerignoux: can we keep the file_name == main ClassName ?

nyuki/web.py
```diff
@@ -0,0 +1,187 @@
...
+    WebHandler.READY_CALLBACK = decorated
+    return func
+
+
+class WebHandler(Service):
+
+    CONF_SCHEMA = {
+        "type": "object",
+        "properties": {
+            "websocket": {
```
lerignoux: Rather also keep config name == file name == class name ?

nyuki/web.py
```diff
@@ -0,0 +1,187 @@
...
+    async def stop(self):
+        if self.server:
+            for ws in self.server.websockets:
+                ws.close()
+            self.server.close()
+            await self.server.wait_closed()
+
+    def new_token(self):
+        """
+        Random token using 20 digits/lowercases
```
lerignoux: 30 ?

nyuki/web.py
```diff
@@ -0,0 +1,187 @@
...
+        for websocket in self.server.websockets:
+            await websocket.send(data)
+
+    async def send_ready(self, websocket):
+        """
+        Send the 'ready' message to a client
+        """
+        ready = {
+            'type': 'ready',
+            'keepalive_delay': self.keepalive * 0.8,
```
lerignoux: 0.8 ???

nyuki/web.py
```diff
@@ -0,0 +1,187 @@
...
+            'data': {}
```
lerignoux: not needed, reset below

nyuki/web.py
```diff
@@ -0,0 +1,187 @@
...
+import string
+import websockets
+
+from nyuki.services import Service
+
+
+log = logging.getLogger(__name__)
+
+
+def websocket_ready(func):
```
lerignoux: what is this used for ?

nyuki/web.py
```diff
@@ -0,0 +1,187 @@
...
+            'keepalive timed out'
+        )
+
+    async def _wshandler(self, websocket, path):
+        """
+        Main handler for a newly connected client
+        """
+        match = re.match(r'^\/(?P<token>[a-zA-Z0-9]{30})$', path)
+        if not match:
+            log.debug('token does not match (%s)', path)
```
lerignoux: no token found ? no token pattern ?

nyuki/web.py
```diff
@@ -0,0 +1,187 @@
...
+            except ValidationError:
+                log.debug('Invalid keepalive received: %s', data)
+                continue
+
+            mtype = data['type']
+            if mtype == 'keepalive':
+                handle.cancel()
+                handle = self._schedule_keepalive(websocket, token)
+
+        handle.cancel()
```
lerignoux: we will never reach this place right ?
how does the front closes a websocket, it should be handled and call a break to exit the while True (better set a variable and unset it on disconnection IMO)

nyuki/nyuki.py
```diff
@@ -236,3 +240,13 @@ def get(self, request):
                 })
 
             return Response(body=body)
+
+    @resource(endpoint='/websocket', version='v1')
+    class WebsocketToken:
```
lerignoux: need swagger update

tests/web_test.py
```diff
@@ -0,0 +1,77 @@
...
+        with assert_raises(exceptions.ConnectionClosed):
+            await self.client.recv()
+
+        # Create token using API
+        resp = self.nyuki.WebsocketToken.get(self.nyuki, None)
+        token = json.loads(resp.body.decode())['token']
+        assert_in(token, web._tokens)
+
+        # Good token
+        self.client = await client.connect('ws://localhost:5566/' + token)
```
lerignoux: currently unsecure right ? let's create a task to encrypt the socket transactions

nyuki/web.py
gdraynz: I agree to rename the file to `websocket.py` but no file == class, it's not Java and we can have any number of classes in a file

nyuki/web.py
gdraynz: file renamed `websocket.py`

nyuki/web.py
gdraynz: yup

nyuki/web.py
gdraynz: We send an inferior delay to the clients to avoid calculating it on front end.
Clients will just take this keepalive value and send keepalive packets on this delay

nyuki/web.py
gdraynz: Only if a ready callback is set

nyuki/web.py
gdraynz: Yep, it decorates a method in a nyuki to send data to a client at connection time

nyuki/web.py
gdraynz: uh ?

nyuki/web.py
gdraynz: We do reach this place using the `break` above. Even from a client manual disconnection or a timeout resulting of a keepalive not.

nyuki/nyuki.py
gdraynz: ANNOYING MOMENT NUMBER ONE it must be added in each individual nyuki

tests/web_test.py
gdraynz: We should indeed check it using a frontend websocket client

nyuki/nyuki.py
lerignoux: created NYSWAG-2

nyuki/web.py
lerignoux: deal rename the file

nyuki/web.py
lerignoux: indeed

PR #63, opened 2016-02-04
Add a reporting module wrapping standard messages published by nyukis

gdraynz:
First draft that define the global functional perimeter.
Json validation at emission and reception and publication into the `reports` muc, for now.

### DIFF
--- a/nyuki/bus.py
+++ b/nyuki/bus.py
@@ -241,7 +241,7 @@ class Bus(Service):
         else:
             log.warning('No callback set for direct messages')
 
-    async def publish(self, event):
+    async def publish(self, event, dest=None):
         """
         Send an event in the nyuki's own MUC so that other nyukis that joined
         the MUC can process it.
@@ -259,7 +259,7 @@ class Bus(Service):
         log.debug(">> publishing to '{}': {}".format(self._topic, event))
         msg = self.client.Message()
         msg['type'] = 'groupchat'
-        msg['to'] = self._muc_address(self._topic)
+        msg['to'] = self._muc_address(dest or self._topic)
         msg['body'] = json.dumps(event)
         log.info('Publishing an event to %s', msg['to'])
         msg.send()
@@ -307,7 +307,7 @@ class Bus(Service):
         self._direct_message_callback = None
         log.info('Unsubscribed from direct messages')
 
-    def send_message(self, recipient, data):
+    async def send_message(self, recipient, data):
         """
         Send a direct message to 'recipient'
         """
@@ -316,6 +316,9 @@ class Bus(Service):
             return
         if not isinstance(data, dict):
             raise TypeError('Message must be a dict')
+        if not self._connected.is_set():
+            log.info("Waiting for a connection to direct message '%s'", recipient)
+        await self._connected.wait()
         log.debug(">> direct message to '{}': {}".format(recipient, data))
         body = json.dumps(data)
         self.client.send_message(recipient, body)
--- a/nyuki/nyuki.py
+++ b/nyuki/nyuki.py
@@ -96,7 +96,7 @@ class Nyuki(metaclass=CapabilityHandler):
 
     def _exception_handler(self, loop, context):
         log.debug('Exception context: %s', context)
-        if 'exception' in context:
+        if 'exception' not in context:
             log.warning(context)
             return
 
--- a/nyuki/nyuki.py
+++ b/nyuki/nyuki.py
@@ -11,6 +11,7 @@ from nyuki.commands import get_command_kwargs
 from nyuki.config import get_full_config, write_conf_json, merge_configs
 from nyuki.handlers import CapabilityHandler
 from nyuki.logs import DEFAULT_LOGGING
+from nyuki.reporting import Reporter
 from nyuki.services import ServiceManager
 from nyuki.websocket import WebHandler
 
@@ -71,6 +72,7 @@ class Nyuki(metaclass=CapabilityHandler):
         if self._config.get('websocket') is not None:
             self._services.add('websocket', WebHandler(self))
 
+        self.reporter = None
         self.is_stopping = False
 
     def __getattribute__(self, name):
@@ -106,6 +108,12 @@ class Nyuki(metaclass=CapabilityHandler):
 
         # Start services
         self.loop.run_until_complete(self._services.start())
+
+        if 'bus' in self._services.all:
+            self.reporter = Reporter(self.bus.client.boundjid.user, self.bus)
+            asyncio.ensure_future(self.bus.subscribe(
+                'reports', self._handle_report
+            ))
 
         # Call for setup
         if not asyncio.iscoroutinefunction(self.setup):
@@ -199,6 +207,21 @@ class Nyuki(metaclass=CapabilityHandler):
         Called right before closing the event loop, stopping the Nyuki.
         """
         log.warning('Teardown called, but not overridden')
+
+    async def _handle_report(self, body):
+        try:
+            self.reporter.check_report(body)
+        except ValidationError as ve:
+            log.debug('Report validation failed: %s', ve)
+            return
+        log.info('Valid report received, calling handle_report')
+        await self.handle_report(body)
+
+    async def handle_report(self, body):
+        """
+        Called on reception of a report from another nyuki
+        """
+        log.debug('Report received without callback')
 
     def update_config(self, *new_confs):
         """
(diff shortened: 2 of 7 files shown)


nyuki/reporting.py
```diff
@@ -0,0 +1,137 @@
+from datetime import datetime
```
lerignoux: wouldn't a DB reporting be more adapted (data perenity, ...)

nyuki/nyuki.py
```diff
@@ -102,6 +104,12 @@ def start(self):
         # Start services
         self.loop.run_until_complete(self._services.start())
 
+        if 'bus' in self._services.all:
+            self.reporter = Reporter(self.bus.client.boundjid.user, self.bus)
```
lerignoux: wouldn't a DB reporting be more adapted (data perenity, ...)

tests/reporting_test.py
```diff
@@ -0,0 +1,80 @@
...
+    def test_001_check_type(self):
+        with assert_raises(ValidationError):
+            self.reporter.check_report({
+                'type': 'something',
+                'author': 'test',
+                'date': datetime.utcnow().isoformat(),
+                'data': {
+                    'address': 'test'
+                }
+            })
```
lerignoux: you also need to check failing cases

nyuki/nyuki.py
```diff
@@ -180,6 +188,21 @@ def _validate_config(self, config=None):
...
+            log.debug('Report validation failed: %s', ve)
+            return
+        log.info('Valid report received, calling handle_report')
+        await self.handle_report(body)
+
+    async def handle_report(self, body):
+        """
+        Called on reception of a report from another nyuki
+        """
+        log.debug('Report received without callback')
```
lerignoux: what od you intend to do with that ?
shouldn't it be a specific nyuki that handles it ? not sure it should be part of nyuki

nyuki/reporting.py
```diff
@@ -0,0 +1,137 @@
...
+        'author': {
+            'type': 'string',
+            'minLength': 1
+        },
+        'date': {
+            'type': 'string',
+            'format': 'isoformat'
+        },
+        'data': {'oneOf': [
+            {'$ref': '#/definitions/{}'.format(rtype)} for rtype in ReportTypes.all()
```
lerignoux: why report type is in data and not type ?

nyuki/reporting.py
```diff
@@ -0,0 +1,137 @@
...
+        return [item.name for item in cls]
+
+
+REPORT_SCHEMA = {
+    'type': 'object',
+    'required': ['type', 'author', 'date', 'data'],
+    'properties': {
+        'type': {
+            'type': 'string',
+            'format': 'report_type'
```
lerignoux: shouldn't that be a oneOf ?

nyuki/nyuki.py
gdraynz: It is already tested in the tests below

tests/reporting_test.py
gdraynz: It is already tested in the tests below

nyuki/nyuki.py
gdraynz: Added this to eventually handle reports from any nyuki, as it is a common functionality to every nyuki

nyuki/reporting.py
gdraynz: `data` contains the report data depending of its own type and so its own schema

nyuki/reporting.py
gdraynz: oneOf needs to be used for definitions, we can't directly check that a string if of an exact value in the schema

nyuki/nyuki.py
gdraynz: I don't think adding a database in each nyuki is a good thing

thavel: I agree with @gdraynz.

nyuki/bus.py
```diff
@@ -316,6 +316,9 @@ def send_message(self, recipient, data):
             return
         if not isinstance(data, dict):
             raise TypeError('Message must be a dict')
+        if not self._connected.is_set():
+            log.info("Waiting for a connection to direct message '%s'", recipient)
+        await self._connected.wait()
```
thavel: Should not be in the `if` condition below?

nyuki/nyuki.py
```diff
@@ -30,8 +31,6 @@ class Nyuki(metaclass=CapabilityHandler):
     single-threaded, asynchronous and concurrent-safe environment.
     The core engine of a nyuki implementation is the asyncio event loop
     (a single loop is used for all features).
-    A wrapper is also provide to ease the use of asynchronous calls
-    over the actions nyukis are inteded to do.
```
thavel: :+1:

nyuki/nyuki.py
```diff
@@ -107,6 +107,12 @@ def start(self):
         # Start services
         self.loop.run_until_complete(self._services.start())
 
+        if 'bus' in self._services.all:
+            self.reporter = Reporter(self.bus.client.boundjid.user, self.bus)
+            asyncio.ensure_future(self.bus.subscribe(
+                'reports', self._handle_report
```
thavel: Would be great to make the report channel configurable! Default value to `reports` sound good though.

nyuki/nyuki.py
thavel: I agree with @lerignoux. Let's the nyuki that would handle reports implements its own handlers. See my comment below regarding the channel name.

nyuki/reporting.py
```diff
@@ -0,0 +1,137 @@
...
+                'type': 'string',
+                'minLength': 1
+            }
+        },
+        "additionalProperties": False
+    }
+
+    connection_lost = {
+        'type': 'object',
+        'required': ['address'],
```
thavel: `address`? At least an `uri`, no?

nyuki/reporting.py
```diff
@@ -0,0 +1,137 @@
...
+            'address': {
+                'type': 'string',
+                'minLength': 1
+            }
+        },
+        "additionalProperties": False
+    }
+
+    @classmethod
+    def all(cls):
```
thavel: Is it really useful? Could we just code `if something in ReportTypes`?

nyuki/reporting.py
```diff
@@ -0,0 +1,137 @@
...
+
+        try:
+            self.check_report(report)
+        except ValidationError as ve:
+            log.exception(ve)
+            log.error('Report validation failed: %s', str(ve))
+            return
+
+        log.info('Sending report data: %s', report)
+        await self._publisher.publish(report, 'reports')
```
thavel: See my previous comment on hard coded channel name.

thavel:
On a general level: reports types are strict. `data` content should be more permissive.

nyuki/bus.py
gdraynz: It's a deadlock if you `await` it, a call using `ensure_future` will just put it on hold until the connection is set.

nyuki/nyuki.py
gdraynz: That would means that every call to these reports won't be of any use unless a specific nyuki is up and running.
I figured that any nyuki could listen to these events to eventually do something when, let's say a fellow SMS nyuki loses its connection, another SMS-capable nyuki could hear it and carry on its work...

nyuki/reporting.py
gdraynz: I don't think it works that way :)

nyuki/bus.py
thavel: So the task never ends, but the thread isn't blocked, right?

nyuki/nyuki.py
```diff
@@ -89,11 +90,39 @@ def __getattribute__(self, name):
     def config(self):
         return self._config
 
+    @property
+    def report_channel(self):
+        return self.config.get('report_channel', 'errors')
+
+    def _exception_handler(self, loop, context):
+        log.debug('Exception context: %s', context)
+        if 'exception' in context:
```
thavel: Erh, few nested `if`/`else` may be replaced by stop conditions:
```
if 'exception' not in context:
    log.warning(context)
    return
```

nyuki/nyuki.py
```diff
@@ -107,6 +136,14 @@ def start(self):
         # Start services
         self.loop.run_until_complete(self._services.start())
 
+        if 'bus' in self._services.all:
+            asyncio.ensure_future(self.bus.subscribe(
```
thavel: Still to be removed (aka no default handler), IMO.

nyuki/nyuki.py
```diff
@@ -89,11 +90,39 @@ def __getattribute__(self, name):
...
+                except Exception as e:
+                    log.exception(e)
+                    exc = traceback.format_exc()
+                else:
+                    exc = 'could not retrieve exception'
+            else:
+                log.exception(context['exception'])
+                exc = traceback.format_exc()
+
+            if self.reporter:
```
thavel: Do we _really_ want to report all exceptions?

PR #64, opened 2016-02-05
Default configuration file

gdraynz:


nyuki/config.py
```diff
@@ -53,16 +52,28 @@ def merge_configs(*configs):
     return new_conf
 
 
+def read_default_json(return_json=True):
+    """
+    Read the default configuration file
+    No error check, just break everything is it not available
+    """
+    with open(DEFAULT_CONF_FILE, 'r') as f:
```
Ketouem: A `FileNotFoundError` will be thrown if `default.json` does not exist. Such case should be a deal breaker and stop the nyuki.

tests/nyuki_test.py
```diff
@@ -102,3 +84,23 @@ def test_005b_custom_schema_ok(self):
...
+    def tearDown(self):
+        if os.path.exists('something.json'):
+            os.remove('something.json')
+        if os.path.exists('default.json'):
+            os.remove('default.json')
+
+    @ignore_loop
+    def test_001_copy_default(self):
+        with assert_raises(FileNotFoundError):
+            open('something.json')
```
Ketouem: Stage this stuff in a proper temp folder, there is some tooling inside the standard library using [tempfile](https://docs.python.org/3/library/tempfile.html)

tests/nyuki_test.py
```diff
@@ -102,3 +84,23 @@ def test_005b_custom_schema_ok(self):
...
+class TestNyukiWithConfig(TestCase):
+
+    def tearDown(self):
+        if os.path.exists('something.json'):
+            os.remove('something.json')
+        if os.path.exists('default.json'):
+            os.remove('default.json')
+
+    @ignore_loop
+    def test_001_copy_default(self):
```
Ketouem: Add some more tests to cover the cases:
- missing default.json when loading with no config
- loading with a bad config file

nyuki/config.py
gdraynz: `FileNotFoundError` does indeed raise here and break the nyuki's `__init__`

PR #71, opened 2016-02-25
Allow case insensitive lookup, as it was planned

gdraynz:


nyuki/transform.py
```diff
@@ -285,8 +285,11 @@ class Lookup(_Rule):
     Implements a dead-simple lookup table which perform case sensitive
     (default) or insensitive (icase=True) lookups.
     """
-    def _configure(self, table=None):
-        self.table = table or dict()
+    def _configure(self, table=None, icase=False):
+        self.table = table or {}
```
lerignoux: nope
if you use this twice the same dictionary will be reused ! keep dict()

lerignoux: it does not change anything here, but dict(would be safer) {} is mutable and if it is reused it could be altered (cf last commit in this file)

lerignoux: ok

PR #73, opened 2016-03-02
Improve reporting API

thavel:


nyuki/nyuki.py
```diff
@@ -139,7 +139,6 @@ def start(self):
         self.loop.run_until_complete(self._services.start())
 
         if 'bus' in self._services.all:
-            asyncio.ensure_future(self.bus.subscribe(self.report_channel))
```
lerignoux: why do you remove this ?

thavel: Not removed but _moved_. See `nyuki/reporting.py` at [line 60](https://github.com/optiflows/nyuki/pull/73/files#diff-c64dfac25e642ca357bc71e6d7749f96R60).
This is to handle the channel subscription within the reporting module. Otherwise, we have to subscribe twice: once to be able to publish and a second time to listen on event (with a callback).

lerignoux: yes but then if you have a bus and no reporting it is not done ...

thavel: Hopefully yes. If you don't have the reporting module enabled, you should't have the subscription done anyway.

PR #75, opened 2016-03-07
improve exception handler and add a reporter.exception helper

gdraynz:
```python
try:
  something()
except Exception as exc:
  self.reporter.exception(exc)
```

### DIFF
--- a/nyuki/nyuki.py
+++ b/nyuki/nyuki.py
@@ -103,7 +103,7 @@ class Nyuki(metaclass=CapabilityHandler):
             return
         log.debug('Exception context: %s', context)
 
-        exc = None
+        exc = Exception("could not retrieve exception's traceback")
         if 'future' in context:
             try:
                 context['future'].result()
@@ -111,11 +111,6 @@ class Nyuki(metaclass=CapabilityHandler):
                 exc = e
         else:
             exc = context['exception']
-
-        if not exc:
-            exc = Exception("could not retrieve exception's traceback")
-        else:
-            log.exception(exc)
 
         self.reporter.exception(exc)
 
--- a/nyuki/nyuki.py
+++ b/nyuki/nyuki.py
@@ -3,9 +3,8 @@ import json
 from jsonschema import validate, ValidationError
 import logging
 import logging.config
-from signal import SIGHUP, SIGINT, SIGTERM
-import traceback
 from pijon import Pijon
+from signal import SIGHUP, SIGINT, SIGTERM
 
 from nyuki.bus import Bus
 from nyuki.capabilities import Exposer, Response, resource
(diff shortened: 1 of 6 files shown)


nyuki/nyuki.py
```diff
@@ -103,22 +102,21 @@ def _exception_handler(self, loop, context):
             log.warning(context)
```
lerignoux: Better log something else, the context was already logged in debug.
if ...
    log.warning (no exception in context {context})
else:
    log.debug

nyuki/nyuki.py
```diff
@@ -103,22 +102,21 @@ def _exception_handler(self, loop, context):
...
+            exc = context['exception']
 
-        asyncio.ensure_future(
-            self.reporter.send_report('exception', {'traceback': exc}),
-            loop=loop
-        )
+        if not exc:
+            exc = Exception("could not retrieve exception's traceback")
+        else:
+            log.exception(exc)
```
lerignoux: log anyway, we shouldn't arrive here if there is no accessible exception

tests/reporting_test.py
```diff
@@ -37,7 +37,12 @@ def test_001_check_schema(self):
...
-            await self.reporter.send_report('type', 'nope')
+            await self.reporter.send_report('type', 'not a dict')
         await self.reporter.send_report('type', {'key': 'value'})
         # Troubles patching datetime.utcnow
         eq_(self.publisher.publish.call_count, 1)
+
+    async def test_003_exception(self):
+        await self.reporter.exception(Exception('nope'))
+        # Troubles patching datetime.utcnow
+        eq_(self.publisher.publish.call_count, 1)
```
lerignoux: should check the exception text is shown inside

nyuki/nyuki.py
```diff
@@ -103,22 +102,21 @@ def _exception_handler(self, loop, context):
...
-            log.exception(context['exception'])
-            exc = traceback.format_exc()
+            exc = context['exception']
 
-        asyncio.ensure_future(
-            self.reporter.send_report('exception', {'traceback': exc}),
-            loop=loop
-        )
+        if not exc:
+            exc = Exception("could not retrieve exception's traceback")
```
lerignoux: you could init exc with it so that you don't have to override it

lerignoux: I would you a raise immediately caught too to have a traceback and now where is this code (to avoid looking for the exception content within our lib)

lerignoux: but then why not catch the key error on context['exception'] this would lead straight to there, we are just switching one exception for a "fake" one not usre it helps and it may hide things don't you think so ?

nyuki/reporting.py
```diff
@@ -70,3 +74,13 @@ def check_report(self, report):
...
         await self._publisher.publish(report, self._channel)
+
+    async def exception(self, exc):
+        """
+        Helper to report an exception traceback from its object
+        """
+        traceback = TracebackException.from_exception(exc)
+        await self.send_report(
+            'exception',
+            {'traceback': ''.join(traceback.format())}
```
lerignoux: upon logging an error I only got the following post: [nyuki.bus] >> publishing to 'appreminder': {'error': 'We got an exception mayday'}
Shouldn't the traceback be added ?

nyuki/nyuki.py
gdraynz: context is already logged above, it contains everything we know about the exception that led us here

tests/reporting_test.py
gdraynz: I was not able to mock datetime.utcnow to check the publish call parameters :|

nyuki/nyuki.py
gdraynz: You don't want to raise an exception inside the exception handler. That would just duplicates exception and error logs

nyuki/reporting.py
gdraynz: Done, bus was catching exceptions

PR #82, opened 2016-04-12
Bus persistence (using mongo)

gdraynz:
New bus configuration to enable it:
```json
{
    "bus": {
        "storage": {
            "backend": "mongo",
            "host": "localhost",
            "ttl": 3600 (seconds)
        }
    }
}
```


nyuki/bus/persistence/mongo_backend.py
```diff
@@ -0,0 +1,59 @@
+from datetime import datetime
+import logging
+from motor.motor_asyncio import AsyncIOMotorClient
```
lhavel: Nice find!

nyuki/bus/persistence/persistence.py
```diff
@@ -0,0 +1,45 @@
+import logging
+
+from nyuki.bus.persistence.mongo_backend import MongoBackend
+
+
+log = logging.getLogger(__name__)
+
+
+class BusPersistence(object):
```
lhavel: If you define an API that any events storage backend driver must comply with, you don't need to implement a generic wrapper class such as that one. In my opinion, having a well-defined Python API for storage backend drivers is a good idea. You could implement an abstract metaclass with predefined methods to enforce such an API in any driver.

nyuki/bus/bus.py
```diff
@@ -182,13 +224,21 @@ def _muc_address(self, topic):
                 if topic != self._topic:
                     await self.subscribe(topic, callback)
 
+        # Replay events that have been lost, if any
```
lhavel: Automatically re-send lost events once ready to do it is definitely a good idea.
But I don't feel comfortable with your condition: what happens if the nyuki hasn't/couldn't join yet its MUC?

nyuki/bus/bus.py
```diff
@@ -262,17 +306,24 @@ def _on_failure(self, event):
...
-        log.info('Publishing an event to %s', msg['to'])
-        msg.send()
+
+        if self._persistence and store:
+            asyncio.ensure_future(self._persistence.store(
+                dest or self._topic, msg['body']
+            ))
+
+        if not self._persistence and not self._connected.is_set():
+            log.warning('Waiting for a connection to publish')
```
lhavel: If no storage backend is available we talked about a minimal local in-memory storage to keep track of that events. It would be a kind of default/failover storage backend.

nyuki/bus/bus.py
```diff
@@ -262,17 +306,24 @@ def _on_failure(self, event):
...
+            await self._connected.wait()
```
lhavel: Another way to handle that situation is: store the event in the storage area (local in-memory or persistent) anyway and keep going. The _unpublished_ events shall be published automatically upon re-joining the MUC.
You could stall only when the storage backend is full.

nyuki/bus/bus.py
```diff
@@ -262,17 +306,24 @@ def _on_failure(self, event):
...
+
+        if self._connected.is_set():
+            log.debug(">> publishing to '{}': {}".format(self._topic, event))
+            log.info('Publishing an event to %s', msg['to'])
+            msg.send()
```
lhavel: At that time you're almost sure the event has been published but not successfully published!
What if the event is rejected by prosody (e.g. because the nyuki is not in the room)?

nyuki/bus/bus.py
gdraynz: `self.subscribe` already wait for a connection to the xmpp server to execute

nyuki/bus/bus.py
```diff
@@ -79,18 +86,36 @@ class Bus(Service):
...
                         "type": "string",
                         "minLength": 1
                     },
-                    "host": {"type": "string"},
                     "port": {"type": "integer"},
-                    "muc_domain": {"type": ["string", "null"]}
-                }
+                    "storage": {
+                        "type": "object",
+                        "required": ["backend"],
```
lhavel: As long as we do not support any other storage backend, we'd better make `backend` optional (and default to "mongo" anyway).

nyuki/bus/bus.py
```diff
@@ -101,31 +126,63 @@ def __init__(self, nyuki):
...
     async def start(self, timeout=0):
         self.client.connect()
         if timeout:
             await asyncio.wait_for(self._connected.wait(), timeout)
 
+        if self._persistence:
+            try:
+                await self._persistence.init()
+            except StorageError:
+                log.warning(
```
lhavel: do you want to leave the persistent storage of bus events activated if you could not init the db?

gdraynz: Nope, an exception would be raised above

nyuki/bus/bus.py
```diff
@@ -101,31 +126,63 @@ def __init__(self, nyuki):
...
         self._muc_domain = muc_domain
+        self.client.add_event_handler('groupchat_message', self._on_event)
+        self.client.add_event_handler(
+            'muc::{}::message_error'.format(self._muc_address(self._topic)),
+            self._on_failed_publish
+        )
+
+        # Persistence storage
+        if storage:
+            self._persistence = BusPersistence(**{**storage, 'name': self._topic})
```
lhavel: See my comment about the target configuration. Besides, let's keep our naming rules consistent: `storage` in the config file, `persistent` in the bus object...

nyuki/bus/bus.py
```diff
@@ -288,6 +399,16 @@ def _on_failure(self, event):
             self._callbacks[topic] = callback
```
lhavel: I didn't see that, there's a pitfall here: you can subscribe several times to the same MUC but only the 1st callback will be registered. We should enforce a cleaner approach by either allowing a list of callbacks per topic or making sure that you cannot subscribe more than once to a given topic.

nyuki/bus/bus.py
```diff
@@ -288,6 +399,16 @@ def _on_failure(self, event):
...
         log.info("Subscribed to '{}'".format(topic))
 
+    async def resubscribe(self):
```
lhavel: You call that method from private only methods (`_on_start`, `_on_stream_error`). As an internal and automatic mechanism implemented in the bus layer, it shouldn't be part of its public API.

nyuki/bus/bus.py
```diff
@@ -101,31 +126,63 @@ def __init__(self, nyuki):
...
         self.client.add_event_handler('register', self._on_register)
         self.client.add_event_handler('session_start', self._on_start)
+        self.client.add_event_handler('stream_error', self._on_stream_error)
+
+        # MUCs
         self._topic = self.client.boundjid.user
         self._mucs = self.client.plugin['xep_0045']
         self._muc_domain = muc_domain
+        self.client.add_event_handler('groupchat_message', self._on_event)
+        self.client.add_event_handler(
```
lhavel: Do we get this error when `nyuki-hlseven` is connected to Openfire but not in the Prosody's MUC?

nyuki/bus/bus.py
gdraynz: We stated a while ago that one callback per MUC would be enough, but I can add a warning message just in case

nyuki/bus/bus.py
gdraynz: I don't think so, but I was not able to reproduce the `stream_error` error we got in prod in dev

nyuki/bus/bus.py
lhavel: You could just move the joining attempt into the condition and add a warning:
```
        if topic not in self._callbacks:
            self._mucs.joinMUC(self._muc_address(topic), self._topic)
            self._callbacks[topic] = callback
        else:
            log.warning("already joined '{}', won't rejoin".format(topic))
```

nyuki/bus/bus.py
gdraynz: I'd like to stick with the backend as the way to enable persistence.
In some cases we may just not care about persistence at all so there's no point having the `storage` key in the just to set it has `"enable": false`

nyuki/bus/bus.py
gdraynz: yep, will be renamed

nyuki/bus/bus.py
lhavel: yep but it remains activated

nyuki/bus/bus.py
lhavel: Did you try to have a look at the logs from `nyuki-hlseven` running at CH4V?

nyuki/bus/bus.py
```diff
@@ -229,29 +320,21 @@ def _on_failure(self, event):
...
-            body = json.loads(event['body'])
-        except ValueError:
-            body = {}
-
-        if self._direct_message_callback:
-            log.debug('calling direct message callback')
-            await self._direct_message_callback(efrom, body)
-        else:
-            log.warning('No callback set for direct messages')
+        log.info('Replaying events from %s', since)
```
lhavel: Logging not understandable is `since=None`. Either add a condition before logging or change the wording.

nyuki/bus/bus.py
```diff
@@ -229,29 +320,21 @@ def _on_failure(self, event):
...
+        tasks = list()
+        events = await self._persistence.retrieve(since, status)
```
lhavel: What happens if you have `failed` and `pending` events in the queue and you want to replay each of them?
For instance you could fail to publish an event to the MUC (event' status `failed`) and get disconnected right after that (further events marked as `pending`). How to replay them all?

nyuki/bus/bus.py
gdraynz: ok didn't understand that.
An exception here could simply mean the persistence storage is not accessible at the moment, but could be in the future, so we still consider it if we stated we wanted it in the configuration.

lhavel: yep you're right. I re-thought about it afterwards and think it's better to keep it like that.

nyuki/bus/bus.py
lhavel: This is exactly why it defaults to `false`. Hence you don't need to add the storage key if you don't want data persistence. :))))
I still find it more explicit since we won't support another storage backend before a long time.
Besides, you can temporarily disable data persistence for maintenance without removing your config.

nyuki/bus/bus.py
```diff
@@ -202,14 +265,42 @@ def _on_failure(self, event):
...
 
+    async def _on_stream_error(self, event):
+        """
+        Received on stream_error from prosody
+        (TODO: event informations unknown)
+        """
+        log.error('Received unexpected stream error, resubscribing')
+        log.debug('event dump: %s', event)
+        await self.resubscribe()
+        if event['id'] in self._publish_futures:
```
lhavel: Are you sure the `id` you get from the `stream-error` event is the same as the one you stored on publish?

nyuki/bus/bus.py
```diff
@@ -262,17 +345,45 @@ def _on_failure(self, event):
...
+                    status = EventStatus.FAILED
+                    self._publish_futures[uid] = asyncio.Future()
+                else:
+                    status = EventStatus.SENT
+                    log.info("Event successfully sent to MUC '%s'", msg['to'])
+        else:
+            status = EventStatus.NOT_CONNECTED
+
+        # Once we have a result, store it if needed
+        if self._persistence and await self._persistence.ping() and not replay:
```
lhavel: I'm not sure to get it: why do you conditionally store the event based on the value of `replay`?

nyuki/bus/bus.py
```diff
@@ -262,17 +345,45 @@ def _on_failure(self, event):
...
+                try:
+                    await self._publish_futures[uid]
+                except PublishError:
+                    status = EventStatus.FAILED
+                    self._publish_futures[uid] = asyncio.Future()
+                else:
+                    status = EventStatus.SENT
+                    log.info("Event successfully sent to MUC '%s'", msg['to'])
+        else:
+            status = EventStatus.NOT_CONNECTED
```
lhavel: Wrong naming: an event cannot be `NOT_CONNECTED`. `PENDING` seems more appropriate.

nyuki/bus/bus.py
```diff
@@ -202,14 +265,42 @@ def _on_failure(self, event):
...
+            self._publish_futures[event['id']].set_exception(PublishError)
+
+    async def _on_failed_publish(self, event):
+        """
+        Event received when a publish has failed
+        (publish informations in the event)
+        """
+        muc = event['from'].user
+        log.warning("Was not subscribed to MUC '%s', retrying", muc)
+        if muc == self.client.boundjid.user:
```
lhavel: `self.client.boundjid.user` = `self._topic` :)))

nyuki/bus/bus.py
```diff
@@ -202,14 +265,42 @@ def _on_failure(self, event):
...
+            await self.subscribe(muc)
+        else:
```
lhavel: In which situation would you be able to get an error from the XMPP server if you didn't try to publish an event and you're no longer in the MUC?

nyuki/bus/bus.py
```diff
@@ -202,14 +265,42 @@ def _on_failure(self, event):
...
+        self._publish_futures[event['id']].set_exception(PublishError)
+
     async def _on_event(self, event):
         """
         XMPP event handler when a Nyuki Event has been received.
-        Fire EventReceived with (from, body).
         """
         # ignore events from the nyuki itself
         efrom = event['from'].user
         if efrom == self.client.boundjid.user:
```
lhavel: `self.client.boundjid.user` = `self._topic` :)))

nyuki/bus/bus.py
```diff
@@ -202,14 +265,42 @@ def _on_failure(self, event):
...
+            future = self._publish_futures.get(event['id'])
+            if not efrom:
```
lhavel: ???

nyuki/bus/bus.py
```diff
@@ -288,6 +399,16 @@ def _on_failure(self, event):
             self._callbacks[topic] = callback
         log.info("Subscribed to '{}'".format(topic))
 
+    async def resubscribe(self):
+        """
+        Resubscribe everywhere
+        """
+        if self._muc_domain is not None:
+            await self.subscribe(self._topic)
```
lhavel: Why don't you remove that line and just leave the loop below (without the `if topic != self._topic:` condition)?

nyuki/bus/persistence/events.py
```diff
@@ -0,0 +1,12 @@
+from enum import Enum
+
+
+class EventStatus(Enum):
+
+    FAILED = 'FAILED'
+    NOT_CONNECTED = 'NOT_CONNECTED'
```
lhavel: `PENDING`?

nyuki/bus/persistence/mongo_backend.py
```diff
@@ -0,0 +1,82 @@
...
+        # Get collection for this nyuki
+        self.client = AsyncIOMotorClient(self.host)
+        db = self.client['bus_persistence']
+        self._collection = db[self.name]
+        await self._index_ttl()
+
+    async def _index_ttl(self):
+        # Set a TTL to the documents in this collection
+        await self._collection.create_index(
+            'created_at', expireAfterSeconds=self.ttl*60
```
lhavel: TTL in minutes is too fined grained. TTL in hours would be more usable (use float if we want to test in minutes => 0.5 x 60 x 60 hours = 30 minutes)

nyuki/bus/bus.py
```diff
@@ -262,17 +345,45 @@ def _on_failure(self, event):
...
         msg['type'] = 'groupchat'
         msg['to'] = self._muc_address(dest or self._topic)
         msg['body'] = json.dumps(event)
-        log.info('Publishing an event to %s', msg['to'])
-        msg.send()
+
+        future = asyncio.Future()
+
+        self._publish_futures[uid] = future
+        if not self._persistence or not await self._persistence.ping():
```
lhavel: You'd probably better rely on the persistence layer. For instance, if `self._persistence` is set, we should be able to schedule a coroutine that stores the event at the right place (in-memory or db) asynchronously.

nyuki/bus/bus.py
gdraynz: fixed

tests/bus_test.py
```diff
@@ -99,6 +99,32 @@ def test_001_muc_address(self):
...
+    async def test_001_store(self, store, ping):
+        ping.return_value = True
+        await self.bus.publish({'something': 'something'})
+        eq_(self.bus._persistence.backend.name, 'login')
+        eq_(self.bus._persistence.backend.host, 'localhost')
+        call_dict = store.call_args[0][0]
+        eq_(call_dict['topic'], 'login')
+        eq_(call_dict['message'], '{"something": "something"}')
+        eq_(call_dict['status'], 'NOT_CONNECTED')
+
```
lhavel: Needs to be much more tested...

nyuki/bus/bus.py
gdraynz: But as we require a status for the events we send, we must know the status of the publishment before sending it to mongo to avoid to multiply DB transactions on updates afterwards.

nyuki/bus/persistence/mongo_backend.py
gdraynz: As mentioned by @thavel I'd prefer to keep even seconds on TTLs, even if it force us to store a huge value in our configuration files. It would still be a single value copy/pasted in each of our nyukis. We can have external perspectives from others on that subject.

lhavel: ok, let's keep it in seconds.

nyuki/bus/bus.py
```diff
@@ -100,31 +125,66 @@ def __init__(self, nyuki):
...
+        self.client.add_event_handler('groupchat_message', self._on_event)
+        self.client.add_event_handler(
+            'muc::{}::message_error'.format(self._muc_address(self._topic)),
+            self._on_failed_publish
+        )
+
+        # Persistence storage
+        if persistence:
+            self._persistence = BusPersistence(
+                **{**persistence, 'name': self._topic}
```
thavel: Also work with `BusPersistence(name=self._topic, **persistence)`

nyuki/bus/bus.py
thavel: I agree with @lhavel, the persistency layer ping-method shouldn't be a blocking condition to publish an event. Though, @gdraynz has a valid point: we need the final status before requesting the database. As we discussed, this is also a reason why we need to implement asap a reliable memory cache. Let's address this issue later in the appropriate task.

nyuki/bus/persistence/persistence.py
```diff
@@ -0,0 +1,67 @@
...
+    current asyncio cache which is out of our control. (cf internal NYUKI-59)
+    """
+
+    def __init__(self, backend, **kwargs):
+        # TODO: mongo is the only one yet
+        assert backend == 'mongo'
+        self.backend = MongoBackend(**kwargs)
+
+        # Ensure required methods are available, break if not
+        assert hasattr(self.backend, 'init')
```
thavel: Strongly disagree with the assert calls, especially in the `__init__` method. This is not pythonic, plus there is a tool in the object-oriented paradigm called __interfaces__ (or sometimes __protocols__).
We should make our backends inherit from this... and if we really want to be sure, do only a polymorphic type checking (basically, with `isinstance(self.backend, PersistenceBackend)`).

gdraynz: fixed

thavel:
After functionals tests, there are two things to fix or improve:
* When the bus is back on track, the states store in DB is not updated (they are in `PENDING`).
* Upon bus reconnection, I would advise to wait for a second or so before replaying missed events. If the XMPP server is unreachable for one nyuki, it's would be more likely the same for all of them, therefore we should wait a "safely" period so the other nyukis could have time to reconnect and join the MUC back.

nyuki/nyuki.py
```diff
@@ -299,6 +299,42 @@ def get(self, request):
 
             return Response(self._config)
 
+    @resource(endpoint='/bus/replay', version='v1')
```
thavel: I've the following exception while using this route:
```
hl7_1                | 2016-04-19 13:34:34,611  DEBUG    [nyuki.nyuki] Exception context: {'exception': UnboundLocalError("local variable 'status' referenced before assignment",), 'message': "local variable 'status' referenced before assignment"}
hl7_1                | 2016-04-19 13:34:34,613  ERROR    [nyuki.bus.reporting] Traceback (most recent call last):
hl7_1                |   File "/usr/lib/python3.5/site-packages/nyuki/api.py", line 94, in middleware
hl7_1                |     capa_resp = await capa_handler(request, **request.match_info)
hl7_1                |   File "/usr/lib/python3.5/asyncio/coroutines.py", line 218, in coro
hl7_1                |     res = yield from await_meth()
hl7_1                |   File "/usr/lib/python3.5/site-packages/nyuki/nyuki.py", line 336, in post
hl7_1                |     await self.bus.replay(since, status)
hl7_1                | UnboundLocalError: local variable 'status' referenced before assignment
hl7_1                |
hl7_1                | 2016-04-19 13:34:34,616  DEBUG    [nyuki.bus.reporting] Exception already logged
hl7_1                | 2016-04-19 13:34:34,617  ERROR    [nyuki.api] Error handling request
hl7_1                | Traceback (most recent call last):
hl7_1                |   File "/usr/lib/python3.5/site-packages/aiohttp/server.py", line 262, in start
hl7_1                |     yield from self.handle_request(message, payload)
hl7_1                |   File "/usr/lib/python3.5/site-packages/aiohttp/web.py", line 87, in handle_request
hl7_1                |     resp = yield from handler(request)
hl7_1                |   File "/usr/lib/python3.5/site-packages/nyuki/api.py", line 107, in middleware
hl7_1                |     raise exc
hl7_1                |   File "/usr/lib/python3.5/site-packages/nyuki/api.py", line 94, in middleware
hl7_1                |     capa_resp = await capa_handler(request, **request.match_info)
hl7_1                |   File "/usr/lib/python3.5/asyncio/coroutines.py", line 218, in coro
hl7_1                |     res = yield from await_meth()
hl7_1                |   File "/usr/lib/python3.5/site-packages/nyuki/nyuki.py", line 336, in post
hl7_1                |     await self.bus.replay(since, status)
hl7_1                | UnboundLocalError: local variable 'status' referenced before assignment
hl7_1                | 2016-04-19 13:34:34,622  DEBUG    [nyuki.api.access] 172.17.0.21 - - [19/Apr/2016:13:34:34 +0000] "POST /v1/bus/replay HTTP/1.0" 500 170 "-" "CocoaRestClient/16 CFNetwork/760.4.2 Darwin/15.4.0 (x86_64)"
```

PR #86, opened 2016-04-26
Add in-memory persistence

gdraynz:


nyuki/bus/bus.py
```diff
@@ -394,15 +391,17 @@ def _on_failure(self, event):
         del self._publish_futures[uid]
 
         # Once we have a result, store it if needed
-        if self._persistence and await self._persistence.ping() and not previous_uid:
+        if previous_uid:
+            log.info("Updating stored event '%s' with status '%s'", uid, status)
+            await self._persistence.update(previous_uid, status)
+        elif self._persistence:
```
thavel: Since the in-memory buffer is defined in the the `BusPersistence` class, I guess we'll never use it if no persistence storage is defined in the config file. But we want to.

nyuki/bus/persistence/persistence.py
```diff
@@ -11,27 +14,106 @@ class PersistenceError(Exception):
     pass
 
 
+class FIFOSizedQueue(object):
```
thavel: Why not using `asyncio.queues.Queue` (async FIFO sized queue)?

nyuki/bus/persistence/persistence.py
```diff
@@ -11,27 +14,106 @@ class PersistenceError(Exception):
...
+                await self._empty_last_events()
+                break
+
+            if not self._last_events.list:
+                continue
+
+            await self._empty_last_events()
+
+    async def _empty_last_events(self):
+        if await self.backend.ping():
```
thavel: I would suggest a timeout (could be here or in higher calls), otherwise, if the persistence storage is being stopped just before the nyuki, the nyuki would likely never stop.

nyuki/bus/persistence/persistence.py
```diff
@@ -42,44 +124,76 @@ def __init__(self, backend, **kwargs):
...
         }
+        adding a 'created_at' key.
         """
-        try:
-            return await self.backend.store(event)
-        except Exception as exc:
-            raise PersistenceError from exc
+        event['created_at'] = datetime.utcnow()
+
+        if await self.ping():
```
thavel: Why storing the event straight away? Just place the events in the buffer and let the `_feed_backend ` method do its job.

nyuki/bus/persistence/persistence.py
gdraynz: Because we have to access the whole list of events to `retrieve` and `update`, and there's no easy way to do it using `asyncio.Queue`

thavel: `return list(self._queue)` (within the class context).

nyuki/bus/persistence/persistence.py
```diff
@@ -42,44 +124,76 @@ def __init__(self, backend, **kwargs):
...
-            "message": "json dump"
-        }
+        Update the status of a stored event
         """
-        log.info("Updating stored event '%s' with status '%s'", uid, status)
-        try:
-            return await self.backend.update(uid, status)
-        except Exception as exc:
-            raise PersistenceError from exc
+        if await self.ping():
```
thavel: What about an event that isn't in the storage yet, but in the buffer?

nyuki/bus/persistence/persistence.py
```diff
@@ -42,44 +124,76 @@ def __init__(self, backend, **kwargs):
...
 
     async def retrieve(self, since=None, status=None):
         """
         Must return the list of events stored since the given datetime
         """
-        try:
-            return await self.backend.retrieve(since=since, status=status)
-        except Exception as exc:
-            raise PersistenceError from exc
+        if await self.ping():
```
thavel: You're assuming the events to retrieve _are_ in the storage. What about events in the buffer?

gdraynz: If no backend is accessible (no ping or not set up), it just go to the `filter()` call below that uses the memory

nyuki/bus/persistence/persistence.py
gdraynz: I chose to immediately use the backend it is available, the `_feed` loop is only here to allow a kind of "reconnection" if the connection to mongo is down and up again

nyuki/bus/persistence/persistence.py
gdraynz: done

nyuki/bus/persistence/persistence.py
gdraynz: If the backend is available, the bus throws everything in it, the memory is only here if no backend is set up or if the backend is down.

nyuki/bus/bus.py
gdraynz: fixed

nyuki/bus/persistence/persistence.py
thavel: So you're looking for events to replay either in-memory +or+ in the database (according to the current state). I guess we can find cases where there're events stored in both side.

As a general matter on this buffer implementation, I would recommend to always use it. And then decide to dump its content periodically into the database (if available) or drop oldest items if the buffer is full.

This is a basic CPU hit/miss cache pattern. Not sure we should implement something else...

nyuki/bus/persistence/persistence.py
```diff
@@ -11,75 +14,197 @@ class PersistenceError(Exception):
...
+
+        log.debug('event not found in memory, checking backend')
+
+        if self.backend:
+            async def _ensure_status():
+                while True:
+                    try:
+                        return await self.backend.update(uid, status)
+                    except Exception as exc:
+                        raise PersistenceError from exc
```
thavel: If you raise all exceptions as a `PersistenceError`, your retry mechanism would never work.

PR #91, opened 2016-05-10
Make reporting global

gdraynz:


nyuki/bus/reporting.py
```diff
@@ -10,6 +10,33 @@
 log = logging.getLogger(__name__)
 
 
+_reporter = None
```
thavel: Not a big fan of wrapping every reporter's methods.

Guido [said|https://mail.python.org/pipermail/python-ideas/2012-May/014969.html] something about it:
> There is actually a hack that is occasionally used and recommended: a module can define a class with the desired functionality, and then at the end, replace itself in sys.modules with an instance of that class.

In our example:
```python
sys.modules[__name__] = Reporting()
```
The reporting class should have the same `init()` method as you wrote and a `__getattr__()` to access the singleton attributes.

So we could do:
```python
from nyuki import reporting
reporting.init()
reporting.exception()
```

nyuki/nyuki.py
```diff
@@ -98,6 +98,15 @@ def config(self):
     def report_channel(self):
         return self.config.get('report_channel', 'monitoring')
 
+    @property
+    def reporter(self):
+        """
+        Ensure backwards compatibility
+        TODO: must be removed
```
thavel: Let's maintain backward compatibility until the next (and first) stable version. 👍

nyuki/bus/reporting.py
gdraynz: cool, done

PR #101, opened 2016-06-22
[NYUKI-81] Add a key to template errors.

thavel:


nyuki/workflow/validation.py
```diff
@@ -1,31 +1,72 @@
...
+        """
+        Gives JSON formatted details for a given exception on a task:
+        {
+            "task_id": "<id>",
+            "config_path": "<key.key.key>",
+            "error": "<type>",
+            "info": "<extra>"
+        }
+        """
+        assert isinstance(exc, jsonschema.ValidationError)
```
lhavel: The method is called only on `jsonschema.ValidationError`. This statement is not required.

nyuki/workflow/validation.py
```diff
@@ -1,31 +1,72 @@
...
+        path = list(exc.absolute_path)
+        error = exc.validator
+        value = exc.validator_value
+
+        # Get a proper path for the `required` validation error
+        if error == 'required':
+            cursor = task.get('config', {})
+            for key in path:
+                cursor = cursor[key]
+            for key in value:
```
lhavel: When should we expect a `key` from `value` in `cursor`?

nyuki/workflow/validation.py
```diff
@@ -1,31 +1,72 @@
...
-    {
-        "task_id": "<id>",
-        "config_path": "<key.key.key>",
-        "error": "<type>",
-        "info": "<extra>"
-    }
-    Any other error is available through `message`.
+    Generic error messages for template validation.
     """
+    UNKNOWN = 'Unknown error'
```
lhavel: If this is intended to be error keys, shouldn't we have something more key-style like `unknown-error`, `unknown-task-name`...?

nyuki/workflow/validation.py
```diff
@@ -1,31 +1,72 @@
...
     def as_dict(self):
-        return {'message': self.message, 'details': self.details}
+        return {
+            'key': self.key.name,
+            'message': self.message,
+            'details': self.details
+        }
+
+    @staticmethod
+    def format_details(exc, task):
```
lhavel: An exception class with a `staticmethod` is weird. Why don't you just leave it as a simple function?

nyuki/workflow/validation.py
thavel: In case this is a `required` error, upon schema validation, the lib doesn't add the missing `key` to the `path`. It just returns that something is missing from `path` and it could be one of the `value`.

nyuki/workflow/validation.py
```diff
@@ -40,12 +81,16 @@ def validate(template):
         template.validate()
     except UnknownTaskName as exc:
```
lhavel: Here is raises an error as soon as it encounters an unknown task name. But below you walk through the whole config to gather all schema validation errors.
We must chose what policy to apply:

1. look for all invalid task names and all schema errors
1. raise an exception as soon as an error is encountered.

nyuki/workflow/validation.py
thavel: Done.

PR #104, opened 2016-06-28
[NYUKI-85] Workflow templates listen on their own topics (ID)

thavel:


nyuki/workflow/workflow.py
```diff
@@ -91,6 +91,20 @@ def topics(self):
...
         await self.engine.data_received(data, efrom)
 
+    async def load_template(self, template):
+        """
+        Wrapper to tukio `engine.load`. Parameter is a template object.
+        We want workflow templates to listen to their own topic (unicast).
+        """
+        # Ignore if `topics` is None (broadcast)
+        if template.topics is not None:
+            template.topics += [template.uid]
```
lerignoux: append is more effective

nyuki/workflow/workflow.py
```diff
@@ -91,6 +91,20 @@ def topics(self):
...
+        try:
+            await self.engine.load(template)
+        except Exception as exc:
+            # Means a bad workflow is in database, report it
+            reporting.exception(exc)
```
lerignoux: why ? !
we want to the the exceptions !
any except exception should have at least a log.exception !!!

nyuki/workflow/workflow.py
```diff
@@ -427,35 +438,41 @@ class Workflows:
...
             if not templates:
                 return Response(status=404, body={
                     'error': 'Could not find a suitable template to run'
                 })
 
-            wflow = self.engine.run(
-                WorkflowTemplate.from_dict(templates[0]),
-                request.get('inputs', {})
+            # Create an instance through its own topic
+            instances = await self.engine.data_received(
```
lerignoux: will htat start draft as well ?
I seem to remember we wanted a specific method to start draft in a specific way. is it another one ? deprecated ?

PR #113, opened 2016-07-19
Add schemas and validations for the factory task

gdraynz:


nyuki/workflow/tasks/factory.py
```diff
@@ -1,44 +1,150 @@
...
+                    'type': {'type': 'string', 'enum': [factory_type]},
+                    'rules': {
+                        'type': 'array',
+                        'items': FACTORY_SCHEMAS[factory_type]
+                    }
+                }
+            } for factory_type in FACTORY_SCHEMAS.keys()
+        }
+    }
+    BASE_API_URL = 'http://localhost:5558/v1/workflow'
```
lhavel: Hard-coded that way?

gif-vomi-gif-du-lundi

nyuki/workflow/tasks/factory.py
```diff
@@ -1,44 +1,150 @@
...
+            for ruler in config['rulers']:
+                if ruler['type'] in ['extract', 'sub']:
+                    await self.get_regexes(session, ruler['rules'])
+                elif ruler['type'] == 'lookup':
+                    await self.get_lookups(session, ruler['rules'])
+
+    async def execute(self, event):
         data = event.data
-        converter = Converter.from_dict(self.config)
+        runtime_config = deepcopy(self.config)
```
lhavel: Why do you copy this? It cannot change during execution (tukio already ensures that)

nyuki/workflow/tasks/factory.py
```diff
@@ -1,44 +1,150 @@
+from aiohttp import ClientSession
+from copy import deepcopy
 import logging
-from nyuki.transform import Converter
 from tukio.task import register
 from tukio.task.holder import TaskHolder
 
+from nyuki.transform import Converter
```
lhavel: I think the current `transform.py` module brings unnecessary complexity. We should keep it simpler and maybe re-use only parts of it.

nyuki/transform.py
```diff
@@ -26,149 +26,114 @@ def get(mcs):
...
-            ruler.apply(data)
+        for rule in self.rules:
+            rule.apply(data)
 
 
-class Ruler(object):
+class ConditionBlock(metaclass=_RegisteredRule):
+
+    TYPENAME = 'condition-block'
+    CONDITION_REGEX = re.compile(
```
lhavel: Do you really want to check whether this is a valid condition that way?
I think there's a better, more powerful and more flexible way to deal with conditions.

thavel: I've managed to use Python syntax trees to check the safety of an arithmetic expression:
https://gist.github.com/thavel/83bb53feac2e4e8e1e2a9f7e588c993e

It allows complex expressions, and we can also customize allowed types, comparisons and operators.

nyuki/transform.py
```diff
@@ -26,149 +29,87 @@ def get(mcs):
...
+        as it must always start with 'if' and end with 'elif', 'else'
+        or nothing.
         """
-        for rule in self.rules:
-            rule.apply(data)
-
-
-class _Rule(metaclass=_RegisteredRules):
+        for cond in self._conditions:
+            # If type 'else', apply it and leave
```
lhavel: A condition block that starts with an else is not valid. You should raise an exception in that case.

nyuki/transform.py
```diff
@@ -26,149 +29,87 @@ def get(mcs):
...
+            if cond['type'] == 'else':
+                Converter.from_dict(cond).apply(data)
+                return
+            # Else find the condition and apply it
+            cleaned = self._clean_condition(cond['condition'], data)
+            log.debug('arithmetics: trying %s', cond['condition'])
+            log.critical(cleaned)
```
lhavel: Remove that log?!

nyuki/transform.py
```diff
@@ -26,149 +29,90 @@ def get(mcs):
...
-
-    @property
-    def type(self):
-        return self._rule_cls.TYPENAME
-
-    @property
-    def rules(self):
-        return self._rules
+    def __init__(self, conditions, type=None):
+        if len(conditions) == 0:
```
lhavel: Enhance a little bit more your checkings to ensure it matches an `if` > `elif` (optional) > `else` (optional) pattern.

nyuki/transform.py
```diff
@@ -26,149 +29,101 @@ def get(mcs):
...
+    def __init__(self, conditions, type=None):
+        # Check there is at least one condition
+        if len(conditions) == 0:
+            raise ValueError('Condition block must contain at least one condition')
+        # Check first condition is 'if'
+        if conditions[0]['type'] != 'if':
+            raise TypeError("First condition type must be 'if'")
+        # Check no 'if' or 'else' are in-between
+        if len(conditions) >= 2:
+            if conditions[-1]['type'] == 'if':
```
lhavel: You'd better write `if conditions[-1]['type'] not in ('elif', 'else')`.
Else you can endup having `blop` without raising any exception here.

nyuki/transform.py
```diff
@@ -26,149 +29,101 @@ def get(mcs):
...
+                raise TypeError("Last condition type must not be 'if'")
+            mid_condition_types = [c['type'] for c in conditions[1:-1]]
+            if 'if' in mid_condition_types or 'else' in mid_condition_types:
```
lhavel: Same comment here: you'd better check that you have what you expect instead of testing arbitrary invalid values. Try with `blop` again and that won't raise any exception.

lhavel: For instance something like that:
```
if len(conditions) >= 2 :
    for cond in conditions[1:-1]:
        if cond['type'] != 'elif':
            raise
    if conditions[-1] not in ('elif', 'else'):
        raise
```

nyuki/transform.py
lhavel: Conditions are now checked before evaluation.

PR #116, opened 2016-07-25
Async workflow trigger

gdraynz:


nyuki/workflow/workflow.py
```diff
@@ -471,6 +471,27 @@ class Workflows:
                 return Response(status=400, body={
                     'error': 'Could not start any workflow from this template'
                 })
+
```
lerignoux: this method becomes far too long. I'll split it

lerignoux: done

PR #120, opened 2016-08-03
Rework api resources

gdraynz:
Changes:
* Splited all base resources into separate modules to clean things up (`nyuki/api/`)
* Resources can be put anywhere, but needs to be referenced from the nyuki that is using it.
* `version` is now `versions` and is a list of versions for this resource (always optional).
* The nyuki must now be accessed from the resources using `self.nyuki`.
```python
@resource('/my_resource', versions=['v1', 'v2'])
class MyApiResource:

    async def get(self, request):
        await self.nyuki.bus.publish({})


class MyNyuki(Nyuki):

    HTTP_RESOURCES = Nyuki.HTTP_RESOURCES + [MyApiResource]
```


nyuki/api/capabilities.py
```diff
@@ -0,0 +1,97 @@
+from aiohttp.hdrs import METH_ALL
```
lhavel: The main module of a Python package is often named like the package itself: `api.py`

nyuki/api/capabilities.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+    def decorated(cls):
+        cls.RESOURCE_CLASS = ResourceClass(cls, endpoint, versions, content_type)
+        return cls
+    return decorated
+
+
+class ResourceClass:
+
+    def __init__(self, cls, route, versions, content_type):
```
lhavel: Call it `route` or `endpoint` but don't use both. Technically this is more a `route` than an `endpoint`.

nyuki/api/capabilities.py
```diff
@@ -0,0 +1,97 @@
...
+        self._nyuki = nyuki
+        self._nyuki.register_schema(self.CONF_SCHEMA)
+        self._loop = self._nyuki.loop or asyncio.get_event_loop()
+        self._api = WebServer(self._loop)
+        self.host = None
+        self.port = None
+
+    @property
+    def capabilities(self):
+        return self.ENDPOINTS
```
lhavel: I guess this should be `self._nyuki.ENDPOINTS`

nyuki/api/webserver.py
```diff
@@ -62,7 +40,7 @@ def _get_content_type(self, kwargs):
         return kwargs.get('content_type') or kwargs.get('headers', {}).get('content_type')
 
 
-class Api(object):
+class WebServer(object):
```
lhavel: Already implicitely inherits from `object` (just like the other classes you wrote).

nyuki/api/webserver.py
```diff
@@ -121,10 +99,9 @@ def router(self):
     Transform the request data to be passed through a capability and
     convert the result into a web response.
     """
-
     async def middleware(request):
         # Ensure a content-type check is necessary
-        if hasattr(capa_handler, 'CONTENT_TYPE') and await request.text():
+        if getattr(capa_handler, 'CONTENT_TYPE', None) and await request.text():
```
lhavel: Why did you change `hasattr` to `getattr`?

nyuki/nyuki.py
```diff
@@ -1,28 +1,27 @@
...
+from .logs import DEFAULT_LOGGING
+from .services import ServiceManager
+from .websocket import WebHandler
 
 
 log = logging.getLogger(__name__)
 
 
-class Nyuki(metaclass=CapabilityHandler):
+class Nyuki:
```
lhavel: I like that metaclass removed! +1

nyuki/nyuki.py
```diff
@@ -41,6 +40,14 @@ class Nyuki(metaclass=CapabilityHandler):
         "type": "object",
         "required": ["log"]
     }
+    # API endpoints
+    ENDPOINTS = [
```
lhavel: See my comment about the naming rules.

lhavel: Do you really need to redefine the whole list when you create a new nyuki? If yes, we should find a better way to do it.

nyuki/api/webserver.py
gdraynz: `content-type` can be set to `None` to avoid any content-type verification upon receiving a request (useful when we don't care about any data we receive)

nyuki/nyuki.py
gdraynz: In `WorkflowNyuki` we use this method to append resources from the base nyuki and others:
```python
HTTP_RESOURCES = Nyuki.HTTP_RESOURCES + [MyOtherResource, ]
```

nyuki/api/capabilities.py
gdraynz: Done, there was indeed a confusion around naming

lerignoux:
better to use extends than +

lerignoux:
if you want workflow capabilities HTTP_RESOURCES = WorkflowNyuki.HTTP_RESOURCES must be used !!

PR #124, opened 2016-08-16
Change lookup table representation using a list

gdraynz:


nyuki/workflow/api/factory.py
```diff
@@ -243,20 +271,22 @@ class ApiFactoryLookupCSV:
...
+        filename = '{}.csv'.format(lookup['title'])
+        filename = re.sub(r'[ ,]', '_', filename)
+
         # Write CSV
         iocsv = StringIO()
         writer = csv.writer(iocsv, delimiter=',')
-        for key, value in lookup['table'].items():
-            writer.writerow([key, value])
+        for pair in lookup['table']:
+            writer.writerow([pair['value'], pair['replace']])
```
lerignoux: let's Use a DictWritter it's easier since you have a dict in input.

nyuki/workflow/api/factory.py
```diff
@@ -154,13 +178,13 @@ class ApiFactoryLookups:
...
 
-        table = {}
+        table = []
         for row in reader:
             if len(row) != 2:
                 return Response(status=400, body={
                     'error': 'Lookup CSV row length must be 2'
                 })
-            table[row[0]] = row[1]
+            table.append({'value': row[0], 'replace': row[1]})
```
lerignoux: let's use a dict reader it matches our new usage

nyuki/workflow/api/factory.py
```diff
@@ -148,19 +175,12 @@ class ApiFactoryLookups:
...
             log.info('CSV header found: %s', header)
-
-        table = {}
-        for row in reader:
-            if len(row) != 2:
-                return Response(status=400, body={
-                    'error': 'Lookup CSV row length must be 2'
-                })
-            table[row[0]] = row[1]
+        table = [row for row in reader]
```
lerignoux: list(reader) does not work ?

PR #126, opened 2016-08-17
update jsonschema of join task

gdraynz:


nyuki/workflow/tasks/__init__.py
```diff
@@ -12,19 +12,22 @@
...
-            'maxItems': 64,
-            'uniqueItems': True,
-            'items': {
-                'type': 'string',
-                'maxLength': 1024
-            }
+            'anyOf': [
+                {
+                    'type': 'array',
+                    'minItems': 2,
```
lerignoux: why not min 1 ? we could wait for only one parent

nyuki/workflow/tasks/__init__.py
```diff
@@ -12,19 +12,22 @@
...
+                    'maxItems': 64,
+                    'uniqueItems': True,
+                    'items': {
+                        'type': 'string',
```
lerignoux: shouldn't items be a ref to TASKID_SCHEMA ?

nyuki/workflow/tasks/__init__.py
gdraynz: If we have to wait for only one task, there's no point having this task at all.
This is a join, used to gather results of multiple tasks.

nyuki/workflow/tasks/__init__.py
gdraynz: Right

PR #127, opened 2016-08-22
Add generic bus publication task 'notify'

gdraynz:


nyuki/workflow/tasks/notify.py
```diff
@@ -0,0 +1,21 @@
...
+class NotifyTask(TaskHolder):
+
+    """
+    A generic task to simply publish data to the bus.
+    Might trigger linked workflows across the application.
+    """
+
+    async def execute(self, event):
+        log.info('Notifying data to the bus')
+        await runtime.bus.publish(event.data)
```
lerignoux: problem if some events are not serialisable.

For instance in twilio the callback contains a future that is awaited to reply to twilio.
To solve that we may need to send twilio a redirect verb towards the following workflow ID. Twilio will thus call us back on another id and follow on it's tasks

gdraynz: There was already a task serializing every data sent to the bus, no problem here.

PR #134, opened 2016-08-30
csv export latin1

lerignoux:
Back PR for WIWORK-130


nyuki/workflow/api/factory.py
```diff
@@ -263,6 +263,7 @@ class ApiFactoryLookupCSV:
         """
         Return the lookup table for id `lookup_id`
         """
+        encoding = 'ISO-8859-1'
```
gdraynz: Not used

PR #139, opened 2016-09-12
Store finished workflow instances, add routes /history /history/{exec_id}

gdraynz:


nyuki/workflow/workflow.py
```diff
@@ -30,15 +34,20 @@ class BadRequestError(Exception):
...
     """
-    if isinstance(obj, datetime):
-        return obj.isoformat()
-    if isinstance(obj, Workflow):
-        return obj.report()
-    return 'Internal server data'
+    types = [dict, list, tuple, str, int, float, bool, type(None), datetime]
+    if type(obj) not in types:
+        obj = 'Internal server data: {}'.format(type(obj))
+    elif isinstance(obj, dict):
```
lerignoux: returning generators for list and dict would be better.

PR #152, opened 2016-10-03
Checking if the nyuki has a default policy when creating a new template

lerignoux:


nyuki/workflow/api/templates.py
```diff
@@ -105,6 +105,9 @@ class ApiTemplates(_TemplateResource):
                     'error': 'draft already exists'
                 })
 
+        if hasattr(self.nyuki, 'default_policy') and 'policy' not in request:
+            request['policy'] = self.nyuki.default_policy
```
thavel: You're using an instance attribute, I would have go for a class attribute since it's not supposed to change at runtime. Moreover, a initial assign is good for code maintainability.
For instance:
```python
class WorkflowNyuki(Nyuki):
    DEFAULT_POLICY = None
```

lerignoux: Was not a big fan of adding a only workflow related variable nito nyuki but if you think it's better I changed it

PR #157, opened 2016-10-13
[NYUKI-121] Reports of the factory task.

thavel:


nyuki/workflow/tasks/factory.py
```diff
@@ -138,9 +143,13 @@ def __init__(self, config):
...
         log.debug('Full factory config: %s', runtime_config)
         converter = Converter.from_dict(runtime_config)
-        log.debug('Before convertion: %s', data)
-        converter.apply(data)
-        log.debug('After convertion: %s', data)
+        diff = converter.apply(data)
+
+        task = asyncio.Task.current_task()
+        task.dispatch_progress(diff)
+        self._diff = diff
```
lerignoux: why not do directly self._diff = converter.apply(data)

PR #158, opened 2016-10-14
adding report in task selector

lerignoux:


nyuki/workflow/tasks/task_selector.py
```diff
@@ -20,18 +20,28 @@ class TaskConditionBlock(ConditionBlock):
...
+        self._selected = None
 
     def condition_validated(self, rules, data):
         """
         Set next workflow tasks upon validating a condition.
         """
         if rules:
-            self._workflow.set_next_tasks(rules[0]['tasks'])
+            self._workflow.set_next_tasks(rules[0].get('tasks', []))
+            self._selected = rules[0].get('tasks', [])
```
gdraynz: can do this:
```python
self._selected = rules[0].get('tasks', [])
self._workflow.set_next_tasks(self._selected)
```

nyuki/workflow/tasks/task_selector.py
```diff
@@ -20,18 +20,28 @@ class TaskConditionBlock(ConditionBlock):
...
 
 @register('task_selector', 'execute')
 class TaskSelector(TaskHolder):
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._selected = None
+
     SCHEMA = generate_schema(tasks={
         'type': 'object',
```
gdraynz: Add a required key if you're expecting `tasks` to be always given
```
'required': ['tasks']
```

PR #160, opened 2016-10-19
[NYUKI-129] Implement trigger_workflow reports.

thavel:


nyuki/workflow/tasks/trigger_workflow.py
```diff
@@ -109,13 +129,26 @@ def __init__(self, config):
...
-            await asyncio.wait_for(self.async_future, self.timeout)
-            log.info('Instance %s@%s is done', instance, self.nyuki_api)
+            try:
+                await asyncio.wait_for(self.async_future, self.timeout)
+                self._triggered['status'] = Status.DONE.value
+                log.info('Instance %s@%s is done', instance, self.nyuki_api)
+            except asyncio.TimeoutError:
+                self._triggered['status'] = Status.TIMEOUT.value
+                log.info('Instance %s@%s has timeouted', instance, self.nyuki_api)
+            self._task.dispatch_progress(self.report())
```
gdraynz: We should dispatch a cherry-pick of what changed inside the overall reporting data, as we do in every other task.
```json
{
    "status": "new_status"
}
```

thavel: Hm, ok. Let's do the same for all tasks.
Though, I do not agree with this decision.

PR #162, opened 2016-10-31
Multitenant exp

lerignoux:
For review purposes do not merge this experimentation !


nyuki/workflow/tasks/trigger_workflow.py
```diff
@@ -98,6 +98,10 @@ def report(self):
             'Referer': '{}/{}'.format(runtime.bus.name, workflow.uid)
         }
 
+        # 'organization' attribute added by WorkflowNyuki
+        if self.workflow.organization is not None:
```
gdraynz: We agreed on not adding an organization parameter to every task to keep things clean and the organization context seamless throughout the entire user session.
A workflow created on one organization must never be used with any resource of another one.

PR #165, opened 2016-11-21
Rework generic factory schema generator

gdraynz:


nyuki/workflow/tasks/utils/selectors.py
```diff
@@ -1,13 +1,14 @@
-def generate_schema(properties={}, **definitions):
+def generate_factory_schema(schema):
```
thavel: I really don't like this new method signature.

Why not give a "base" schema (optional) and the definition of the rule contents? Something like that:
```python
generate_factory_schema(self, base={'type': 'object'}, **definition)
```
where `definition` can contain a single named dict.

The method would simply add to the base schema:
* Mandatory factory `definitions` (`extract`, `lookup`, etc).
* The generated `rules` into `properties`.
* The `required` property `rules`.

Plus, it doesn't require to refactor all factory-like task schemas.

nyuki/workflow/tasks/utils/selectors.py
```diff
@@ -1,57 +1,56 @@
-def generate_schema(properties={}, **definitions):
+def generate_factory_schema(schema, **rules):
```
thavel: You can maybe set `schema` as an optional parameter with a default value `{'type': 'object'}` which will suit in most cases.

PR #166, opened 2016-11-23
Awaiting trigger workflow by default in trigger workflow task

lerignoux:


nyuki/workflow/tasks/trigger_workflow.py
```diff
@@ -32,7 +32,7 @@ class TriggerWorkflowTask(TaskHolder):
             'nyuki_api': {'type': 'string', 'description': 'nyuki_api'},
             'template': {'type': 'string', 'description': 'template_id'},
             'draft': {'type': 'boolean'},
-            'await_completion': {'type': 'boolean'},
+            'await_completion': {'type': 'boolean', 'default': True},
```
gdraynz: `self.config.get('await_completion')` should also default to True

PR #179, opened 2016-12-06
Sent a count of returned workflow in /workflow/history

gdraynz:


nyuki/workflow/api/README.md
```diff
@@ -0,0 +1,48 @@
...
+## `?full` values
+
+|Value|Effect|
+|-----|------|
+|`0` (default)|Show only basic informations|
+|`1`|Show the entire graphs and tasks exec|
+
+## `?search=something-in-the-title`
+
+## `?since=isoformat-date`
```
lerignoux: let's show an example of an isoformat date

nyuki/workflow/api/utils.py
```diff
@@ -0,0 +1,5 @@
+async def index(collection, *args, **kwargs):
```
lerignoux: I don't really see the use of such an util

nyuki/workflow/api/workflows.py
```diff
@@ -29,6 +34,96 @@ def serialize_wflow_exec(obj):
...
+        asyncio.ensure_future(index(self._instances, 'exec.id', unique=True))
+        asyncio.ensure_future(index(self._instances, 'exec.state'))
+        asyncio.ensure_future(index(self._instances, 'exec.requester'))
+        # Search and sorting indexes
+        asyncio.ensure_future(index(self._instances, [('title', ASCENDING)]))
+        asyncio.ensure_future(index(self._instances, [('title', DESCENDING)]))
+        asyncio.ensure_future(index(self._instances, [('exec.start', ASCENDING)]))
+        asyncio.ensure_future(index(self._instances, [('exec.start', DESCENDING)]))
+        asyncio.ensure_future(index(self._instances, [('exec.end', ASCENDING)]))
+        asyncio.ensure_future(index(self._instances, [('exec.end', DESCENDING)]))
```
lerignoux: these are too many indexes IMO.
I also doubt that both ascending and descending are needed since they are leaf elements.

nyuki/workflow/api/workflows.py
```diff
@@ -221,17 +221,14 @@ class ApiWorkflowsHistory:
                     'error': 'Limit must be an int'
                 })
         try:
-            history = await self.nyuki.storage.instances.get(
+            count, history = await self.nyuki.storage.instances.get(
```
lerignoux: isn't there a problem ? the method is not changed

nyuki/workflow/api/workflows.py
```diff
@@ -29,6 +33,96 @@ def serialize_wflow_exec(obj):
...
+        else:
+            fields = {'_id': 0}
+
+        cursor = self._instances.find(query, fields)
+        # Count total results regardless of limit/offset
+        count = await cursor.count()
+
+        # Set offset and limit
+        if isinstance(offset, int) and offset >= 0:
+            cursor.skip(offset)
```
lerignoux: youskip before the order. I think we should order first otherwise the skip may will just skip randomly (seemingly randomly in the front)

nyuki/workflow/api/workflows.py
```diff
@@ -80,7 +81,7 @@ def __init__(self, instances_collection):
         if isinstance(state, Enum):
             query['exec.state'] = state.value
         if root is True:
-            query['exec.requester'] = None
+            query['exec.requester'] = {'$not': re.compile(r'^nyuki://.*')}
```
lerignoux: is it usefull to compile here ? either compile and store as a variable or just don't compile. I think we are recompiling at every query no ?
also I don't think compile helps for mongo since the most probable is that the string is sent to mongo.

gdraynz: Pymongo automatically converts a `re` object to a query for mongo. It helps by avoiding coding a query not easily writable in python.
But I will compile it in another place.

PR #182, opened 2016-12-09
Better condition regex

gdraynz:


nyuki/utils/evaluate.py
```diff
@@ -80,8 +80,9 @@ def _clean_condition(self, condition, data):
         Format the condition string (as eval-compliant code).
         nb: variable replacement should be `@variable_name` formatted.
         """
+        # /!\ This regex forbids the use of ' and " in a string
         match = re.findall(
-            r'(?:( and| or) +)?\((.*) (~|[=<>!]=?|not in|not|in) (.*)\)',
+            r' *(and|or)? *\( *(@\S*|true|false|[\"\'][^\'\"]*[\'\"]|\d+) +([=<>!]=?|not in|in) +(@\S*|true|false|\d+|[\"\'][^\'\"]*[\'\"]) *\)',
```
thavel: Typo `true` and `false` should be `True` and `False`.
Please also add `None`.

PR #187, opened 2017-01-03
Add a full=1 option in templates route

gdraynz:


nyuki/workflow/api/templates.py
```diff
@@ -260,8 +265,9 @@ class ApiTemplates(_TemplateResource):
         """
         try:
             templates = await self.nyuki.storage.templates.get_all(
-                latest=request.GET.get('latest') in ['true', 'True'],
-                draft=request.GET.get('draft') in ['true', 'True'],
+                full=(request.GET.get('full') == '1'),
```
thavel: Why using integer as boolean? It's really not user-friendly.
If it's a parsing issue, your still can use:
```python
from distutils.util import strtobool
strtobool('true')  # true, True, yes, y, 1
```

gdraynz: This is not supposed to be user-friendly, but propose a simple API.
I'm not fond of allowing multiple values for one boolean parameter and it is already implemented that way for the template history.

thavel: Ok, if it's already implemented this way somewhere else.

PR #188, opened 2017-01-09
websocket rework

gdraynz:
to merge along WIWORK-335. We don't want to break nyuki lib mounting for too long.


nyuki/websocket.py
```diff
@@ -1,62 +1,110 @@
...
+            asyncio.ensure_future(self.remove_client(client, 1001, 'server closing'))
+            for client in self._clients
+        ]
+        if tasks:
+            await asyncio.wait(tasks)
+        self._clients = []
 
-    WebHandler.CLOSE_CALLBACK = decorated
-    return func
+    async def add_client(self, client):
```
lerignoux: shouldn't we use a set and ensure a client cannot be added twice ?

nyuki/websocket.py
```diff
@@ -1,62 +1,110 @@
...
+    async def remove_client(self, client, code=None, reason=None):
+        if client not in self._clients:
+            return
+        if code is None:
+            code = 1000
+        if reason is None:
+            reason = 'connection closed normally'
+        await self.close(client)
+        self._clients.remove(client)
+        await client.close(code, reason)
```
lerignoux: we should await the client close before removing it from _clients, otherwise we may loose the link if the client does not close successfully

nyuki/websocket.py
```diff
@@ -99,130 +133,37 @@ def serializer(self, serializer):
...
+        try:
+            resource = self.RESOURCES[path]
+        except KeyError:
+            await websocket.close(4004, 'resource not found')
             return
 
-        log.info('Connection from token: %s', token)
-        handle = self._schedule_keepalive(websocket, token)
-        await self._send_ready(websocket, token)
+        handle = self._schedule_keepalive(resource, websocket)
```
lerignoux: shouldn't we add the client before planning the keepalive ? (duplicated entry, error during addition ...)

nyuki/websocket.py
```diff
@@ -248,9 +189,8 @@ def _schedule_keepalive(self, websocket, token):
             mtype = data['type']
```
lerignoux: we mandatorily get a keepalive here ? I think the  KEEPALIVE_SCHEMA should be a WEBSOCKET_DATA_SCHEMA instead insn't it ?

nyuki/websocket.py
```diff
@@ -248,9 +189,8 @@ def _schedule_keepalive(self, websocket, token):
...
             if mtype == 'keepalive':
                 handle.cancel()
-                handle = self._schedule_keepalive(websocket, token)
+                handle = self._schedule_keepalive(resource, websocket)
```
lerignoux: shouldn't there be a continue here ? we don't want to cancel and close on keepalive isn't it ?

tests/websocket_test.py
```diff
@@ -1,97 +1,165 @@
...
+        tasks = [
+            asyncio.ensure_future(websockets.connect(
+                'ws://localhost:5559/some/url',
+                extra_headers={'X-Header': 'some header'}
+            ))
+            for _ in range(0, 500)
+        ]
+
+        self.loop.run_until_complete(asyncio.wait(tasks))
+        eq_(len(res._clients), 500)
```
lerignoux: we could try to read two of them to ensure data is received by them all

nyuki/websocket.py
```diff
@@ -1,62 +1,110 @@
...
+            'type': 'ready',
+            'keepalive_delay': self.KEEPALIVE * 0.8,
+            'data': await self.ready(client) or {}
+        }
+        await client.send(json.dumps(ready, default=self._serializer))
+        self._clients.append(client)
 
+    async def remove_client(self, client, code=None, reason=None):
+        if client not in self._clients:
+            return
```
gdraynz: We can remove a client from many places, it will happen. (timeouts, connection lost...)
It is expected.

nyuki/websocket.py
```diff
@@ -1,62 +1,110 @@
...
+        if code is None:
+            code = 1000
+        if reason is None:
+            reason = 'connection closed normally'
```
gdraynz: Like a 200 for HTTP. Everything went fine and the connection closed.

nyuki/websocket.py
```diff
@@ -1,62 +1,110 @@
...
+        log.debug(
+            'Sending data of length %s to %s clients',
+            len(data), len(self._clients)
+        )
+        await asyncio.wait(tasks, timeout=timeout)
+
+    async def ready(self, client):
+        """
+        Called on a new client connection.
+        """
```
gdraynz: There's not necessarily something to say here...

nyuki/websocket.py
```diff
@@ -1,62 +1,110 @@
...
+
+    async def close(self, client):
+        """
+        Called on a client disconnection.
+        """
```
gdraynz: There's not necessarily something to say here...

nyuki/websocket.py
gdraynz: We didn't talk about receiving data from clients for now, It can be done further on.
For now it's only keepalive pings

nyuki/websocket.py
gdraynz: a `continue` at the end of a loop does nothing

nyuki/websocket.py
```diff
@@ -1,62 +1,110 @@
...
+        if tasks:
+            await asyncio.wait(tasks)
+        self._clients = []
 
-    WebHandler.CLOSE_CALLBACK = decorated
-    return func
+    async def add_client(self, client):
+        ready = {
+            'type': 'ready',
+            'keepalive_delay': self.KEEPALIVE * 0.8,
```
gdraynz: This is for the client to use this value as keepalive delay for itself, maybe we should send the exact delay and let the client to the math ?

nyuki/websocket.py
gdraynz: A client is not added twice.
It always comes from one websocket connection and is timed-out/removed if the connection has been terminated.

nyuki/workflow/workflow.py
```diff
@@ -99,6 +111,28 @@ def report(self):
...
+
+    def __init__(self, nyuki, *args, **kwargs):
+        super().__init__(*args, serializer=serialize_wflow_exec, **kwargs)
+        self.nyuki = nyuki
+
+    async def ready(self, client):
+        running = []
+        for workflow in self.nyuki.running_workflows.values():
+            report = workflow.report()
+            del report['graph']
```
gdraynz: the `workflow.report()` I use here creates a dictionary, no need to copy it

nyuki/websocket.py
lerignoux: then 'connection closed properly' would be a better message

nyuki/websocket.py
lerignoux: oh indeed ok

nyuki/workflow/workflow.py
lerignoux: ok but not a big fan, better safe than sorry IMO

tests/websocket_test.py
```diff
@@ -1,97 +1,180 @@
...
+        self.loop.run_until_complete(keepalive())
+        eq_(len(res._clients), 1)
+
+        # Wait for timeout
+        with assert_raises(websockets.ConnectionClosed) as ctx:
+            self.loop.run_until_complete(conn.recv())
+        eq_(ctx.exception.code, 4008)
+        eq_(len(res._clients), 0)
+
+    def test_003_multiple_clients(self):
```
lerignoux: this test takes 3s !! it's far too much. let's get down to 50 we don't need to test that many connections

lerignoux:
to merge along WIWORk-335

nyuki/websocket.py
```diff
@@ -1,62 +1,111 @@
...
+        if tasks:
+            await asyncio.wait(tasks)
+        self._clients = []
 
-    WebHandler.CLOSE_CALLBACK = decorated
-    return func
+    async def add_client(self, client):
+        ready = {
+            'type': 'ready',
+            'keepalive_delay': self.KEEPALIVE,
```
RudolfFrederiksen: a keep alive value of 60 is too long and the websocket keep being closed.
reducing this `keepalive_delay` value from 60 to 48 client (I ignore server conf for the moment), fixed my issue.
the front end used this parameter send by the server to send pings at regular interval to keep the connection alive. 

We may need to get back to that previous value in the backend.

gdraynz: I decided to let the clients adjust their own keepalive instead of dividing it server-side. Simply take like 80% of the returned value and it will be ok.

RudolfFrederiksen: I'm fine with that.

nyuki/websocket.py
```diff
@@ -99,130 +134,37 @@ def serializer(self, serializer):
...
-            log.debug("Unknown token '%s'", token)
-            websocket.close()
-            return
-        elif self.clients[token] is not None:
-            log.debug("Token already in use: '%s'", token)
-            websocket.close()
+        try:
+            resource = self.RESOURCES[path]
+        except KeyError:
+            await websocket.close(4004, 'resource not found')
```
RudolfFrederiksen: I use this value in the front end for error checking.
Does this code has a particular meaning or is it `404` mistyped ?
My guess would be the latter but I'd like to have a confirmation of that.

thavel: WS error codes are often integers > 1000.
https://tools.ietf.org/html/rfc6455#section-7.4.2

PR #190, opened 2017-01-24
[NYPI-70] Bad regex config fail

thavel:
@lerignoux's missing PR

### DIFF
--- a/nyuki/utils/transform.py
+++ b/nyuki/utils/transform.py
@@ -185,7 +185,10 @@ class _Rule(metaclass=_RegisteredRule):
         def wrapper(self, data):
             # Handle data through a traceable dict
             tracker = TraceableDict(data)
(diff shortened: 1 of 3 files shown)


nyuki/utils/transform.py
```diff
@@ -185,7 +189,10 @@ def track_changes(func):
         def wrapper(self, data):
             # Handle data through a traceable dict
             tracker = TraceableDict(data)
-            func(self, tracker)
+            try:
+                func(self, tracker)
+            except RegexpRuleError as exc:
+                return {'type': self.TYPENAME, 'changes': tracker.changes, 'error': 'regexp_rule_error', 'error_details': str(exc)}
```
thavel: PEP-8

lerignoux: yeah yeah fixed !

PR #193, opened 2017-02-01
Removing certificates from xmpp service

lerignoux:


gdraynz:
No need to remove this functionality

PR #194, opened 2017-02-01
Add routes and reporting for pause/resume system

gdraynz:


nyuki/workflow/api/workflows.py
```diff
@@ -260,16 +283,47 @@ class ApiWorkflow(_WorkflowResource):
         except KeyError:
             return Response(status=404)
 
+    async def post(self, request, iid):
```
lerignoux: if we follow the expected changes initiated by @lhavel this method being idempotent it would be a put.
But we need to decide first if we want to switch our current pattern first ...

nyuki/workflow/api/workflows.py
```diff
@@ -182,25 +192,38 @@ class ApiWorkflows(_WorkflowResource):
...
+        else:
+            # Fetch the template from the storage
+            try:
+                templates = await self.nyuki.storage.templates.get(
+                    request['id'],
+                    draft=draft,
+                    with_metadata=True
+                )
+            except AutoReconnect:
+                return Response(status=503)
```
lerignoux: let's include the exception so that we know what is happening

nyuki/workflow/api/workflows.py
thavel: Yes, I'm not a ok with our endpoints either. Although, why would this method be idempotent? "Suspend" or "resume" a workflow twice in a row won't give the same result.

nyuki/workflow/api/workflows.py
thavel: Give the reason (something's wrong with the database connection) is not relevant here. But the status code is (`503`= service temporary unavailable = try again later).

nyuki/workflow/api/workflows.py
gdraynz: A `PUT` must be used when a resource is created from a given payload.
This is just an action on a resource.

nyuki/workflow/api/workflows.py
gdraynz: We already know what's happening, a 503 means the service is temporary unavailable, or "mongo is down".

nyuki/workflow/api/workflows.py
```diff
@@ -114,11 +115,19 @@ def __init__(self, instances_collection):
...
         Insert a finished workflow report into the workflow history.
         """
-        await self._instances.insert(workflow_exec)
+        try:
+            await self._instances.insert(workflow)
+        except DuplicateKeyError:
+            # Ensure a report won't be lost
+            # uuid4 are 8-4-4-4-12 formatted (36 chars), let's add a suffix
+            exec_id = workflow['exec']['id'][:36]
+            suffix = str(uuid4())[:8]
```
gdraynz: I'm not sure adding a suffix is the right solution, we should stay with uuid4 everywhere.
What prevents us from generating another uuid4 for this workflow exec ?

PR #197, opened 2017-02-09
[NYUKI-161] Distributed workflow contexts

thavel:


nyuki/workflow/memory.py
```diff
@@ -0,0 +1,103 @@
...
+        if not config:
+            await self.close()
+            return
+
+        self.service = nyuki_config.get('service')
+        if not self.service:
+            log.error("Can't start Redis: 'service' config key is missing")
+            return
+
+        self.store = await create_reconnecting_redis(
```
gdraynz: From aioredis code:
```python
# NOTE: this is deprecated
create_reconnecting_redis = create_pool
```

nyuki/workflow/memory.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        Setup a shared memory using Redis.
+        """
+        config = nyuki_config.get('redis')
+
+        if not config:
+            await self.close()
+            return
+
+        self.service = nyuki_config.get('service')
```
lerignoux: IMO all memory necessary configuration should be under the same key no ? why split service / redis ?

nyuki/workflow/memory.py
```diff
@@ -0,0 +1,109 @@
...
+    async def clear_report(self, uid):
+        key = self.key('instances', uid)
+        await self.store.delete(key)
+
+    @handle_errors
+    async def write_report(self, report, replace=True):
+        """
+        Store an instance report into shared memory.
+        A simple 'set' is used againts a 'hset' (hash storage), even though the
+        'hset' seems more appropriate because a field in a hash can't have TTL.
```
lerignoux: don't understand why you don't do a hset then ?

nyuki/workflow/memory.py
```diff
@@ -0,0 +1,109 @@
...
+        uid = report['exec']['id']
+        response = await self.store.set(
+            key=self.key('instances', uid),
+            value=pickle.dumps(report),
+            expire=86400,
+            exist=None if replace else False
+        )
+
+        if not response:
+            log.error("Can't share workflow id %s context in memory", uid)
```
lerignoux: key('instance') or uid ?

nyuki/workflow/memory.py
```diff
@@ -0,0 +1,109 @@
...
+
+    @handle_errors
+    async def read_report(self, uid):
+        key = self.key('instances', uid)
+        report = await self.store.get(key)
+        if not report:
+            raise KeyError("Can't find workflow id context %s in memory", uid)
```
lerignoux: key('instance') or uid ?

nyuki/workflow/workflow.py
```diff
@@ -216,21 +218,28 @@ def topics(self):
             ))
         # Enable workflow exec follow-up
         get_broker().register(self.report_workflow, topic=EXEC_TOPIC)
+        await self.memory.setup(self.config)
```
lerignoux: I rather call setup(self.config.get('memory')) or a like to enforce conf partitioning

nyuki/discovery/dns.py
```diff
@@ -1,7 +1,7 @@
 import asyncio
 import logging
-from dns import resolver
-from dns.exception import DNSException
+from aiodns import DNSResolver
```
lerignoux: I guess dns is included in redis packages right ?

requirements.txt
```diff
@@ -6,6 +6,6 @@ pijon>=0.1,<0.2
 # pycares-2.0.0 breaks errno.ECONNRESET
 pycares==1.0.0
 slixmpp>=1.1,<1.2
-tukio>=0.10,<0.11
+tukio>=0.11,<0.12
 websockets>=3.2,<3.3
-dnspython>=1.15,<1.16
+aioredis>=0.3,<0.4
```
lerignoux: missing aiodns dependency

nyuki/discovery/dns.py
thavel: Nope, it's required by slixmpp.

nyuki/workflow/memory.py
```diff
@@ -0,0 +1,109 @@
...
+            await self.store.wait_closed()
+        self.store = None
+
+    async def setup(self, nyuki_config):
+        """
+        Setup a shared memory using Redis.
+        """
+        config = nyuki_config.get('redis')
+
+        if not config:
```
thavel: Adding log to notify that something is _not_ enabled is a waste.

nyuki/workflow/memory.py
thavel: This is not a Redis-specific configuration.

nyuki/workflow/memory.py
thavel: `because a field in a hash can't have TTL`, here is your answer.

nyuki/workflow/memory.py
thavel: Why should I log `key('instance')`, it's relevant. We need to know the workflow execution ID (`uid`).

nyuki/workflow/memory.py
thavel: Same [answer](https://github.com/optiflows/nyuki/pull/197/files#r100792099).

nyuki/workflow/workflow.py
thavel: What's the point? We need multiple information in the config.

requirements.txt
thavel: It's not missing since it's already required by slixmpp.
But you're right, it might be clearer to explicitly add it.

nyuki/workflow/memory.py
```diff
@@ -0,0 +1,109 @@
...
+
+log = logging.getLogger(__name__)
+
+
+def handle_errors(coro):
+    async def wrapper(*args, **kwargs):
+        try:
+            return await coro(*args, **kwargs)
+        except RedisError:
+            log.error("Invalid or badly formatted Redis command")
```
thavel: The whole point of this method is to avoid raising exception and therefore breaking the current execution. So we won't raise it (plus, most of the embedded exceptions are connection-related errors that can be silent since the client reconnects by itself). But you're right, the cause might be interesting log.

nyuki/discovery/dns.py
gdraynz: As we now use it ourselves, I'll add it in the requirements

nyuki/workflow/workflow.py
```diff
@@ -216,21 +218,28 @@ def topics(self):
...
 
     async def reload(self):
         asyncio.ensure_future(self.reload_from_storage())
+        await self.memory.setup(self.config)
 
     async def teardown(self):
         self.global_exec.end()
         if self.engine:
             await self.engine.stop()
+        self.memory.close()
```
gdraynz:
```python
/home/guillaume/src/nyuki/nyuki/workflow/workflow.py:231: RuntimeWarning: coroutine 'Memory.close' was never awaited
  self.memory.close()
```
Also, if you call it properly with `await` you will notice that the redis connection used (AutoConnection) does not allow closing, so you should remove the `self.store.close()` call.

PR #198, opened 2017-02-09
Updating tukio version for workflow stop and resume feature

lerignoux:


lerignoux:
will be done in a later task

PR #203, opened 2017-02-20
High available workflow service.

thavel:


nyuki/workflow/workflow.py
```diff
@@ -341,45 +343,80 @@ def new_workflow(self, template, instance, **kwargs):
...
+                wflow = wflow.decode('utf-8')
+
+                # Get the report shared by the failing instance
+                try:
+                    report = await self.read_report(wflow, ifrom)
+                except KeyError:
+                    log.error("Workflow %s memory has been wiped out", wflow)
+                    break
+
+                shuffle(rescuers)
```
lerignoux: drawing a random id could have been enough, but not a big matter since it will remain small lists.

nyuki/workflow/workflow.py
```diff
@@ -341,45 +343,80 @@ def new_workflow(self, template, instance, **kwargs):
...
+                        'data': report
+                    }
+                    async with aiohttp.ClientSession() as session:
+                        async with session.put(**request) as resp:
+                            if resp.status == 200:
+                                # `ito` rescuer has taken over the workflow
+                                success = True
+                                break
+
+                if not success:
```
gdraynz: You can use this to avoid using a `success` var.
```python
for ito in rescuers:
  something()
  break
else:
  log.error()
  continue
```
The `else` is executed if no `break` occured.

nyuki/workflow/workflow.py
thavel: Actually, we don't want just one ID, but we want to iterate over all ID (randomly) to find a suitable "rescuer".

PR #205, opened 2017-02-22
Improve payload serialization using singledispatch

gdraynz:


nyuki/workflow/workflow.py
```diff
@@ -239,7 +245,7 @@ def new_workflow(self, template, instance, **kwargs):
         """
         wflow = WorkflowInstance(template, instance, **kwargs)
         self.running_workflows[instance.uid] = wflow
-        if self.memory.available:
+        if 'memory' in self._services and self.memory.available:
```
gdraynz: This was breaking if the `memory` service was not configured.

nyuki/services.py
```diff
@@ -25,6 +25,10 @@ def __init__(self, nyuki):
         self.services = dict()
         self._running = False
 
+    def __iter__(self):
+        for name in self.services.keys():
```
thavel: You can also return `self.services` iterator:
```python
def __iter__(self):
    return self.services.__iter__()
```

PR #207, opened 2017-03-01
Add an arithmetic rule for +, -, /, *, U

gdraynz:


nyuki/utils/transform.py
```diff
@@ -401,3 +401,59 @@ def apply(self, data):
...
+
+    def _configure(self, operator, operand1, operand2):
+        self.op = self.OPS[operator]
+        self.operands = (operand1, operand2)
+
+    def _compute_operands(self, data):
+        computed = tuple()
+        for op in self.operands:
+            if isinstance(op, str) and op.startswith('@'):
+                computed += (data[op.split('@')[1]],)
```
lerignoux: I would have gone to a regex to ensure we don't have multiple @ in the name.

nyuki/workflow/tasks/factory.py
```diff
@@ -73,6 +73,23 @@
...
+    'arithmetic': {
+        'type': 'object',
+        'required': ['type', 'fieldname', 'operator', 'operand1', 'operand2'],
+        'properties': {
+            'type': {'type': 'string', 'enum': ['arithmetic']},
+            'fieldname': {'type': 'string', 'minLength': 1},
+            'operator': {'type': 'string', 'enum': ['+', '-', '*', '/', 'union']},
+            'operand1': {'oneOf': [
+                {'type': 'string', 'minLength': 1},
+                {'type': 'integer'},
```
lerignoux: are you sure we won't need float operations ?

lerignoux: do you handle float operands too (in the string parsing I guess)

tests/transform_test.py
```diff
@@ -171,3 +171,36 @@ def test_010a_regexp_error_report(self):
...
+
+        rule = Arithmetic('result', '/', 40, '@int_field_2')
+        rule.apply(data)
+        self.assertEqual(data['result'], 2)
+
+        rule = Arithmetic('result', 'union', '@list_field_1', '@list_field_2')
+        rule.apply(data)
+        self.assertIn({'a': 1}, data['result'])
+        self.assertIn({'b': 2}, data['result'])
+        self.assertIn({'c': 3}, data['result'])
```
lerignoux: let's ensure we don't need float operations. If we do need it let's add some tests

nyuki/utils/transform.py
```diff
@@ -401,3 +401,111 @@ def apply(self, data):
...
+
+
+class Arithmetic(_Rule):
+
+    """
+    Arithmetic rule to add, substrack, multiply and divide fields.
+    """
+
+    OPS = {
+        '+': lambda a, b: a + b,
```
Ketouem: As an alternative you can use the [`operator`](https://docs.python.org/3/library/operator.html#operator.add) module instead of manually defining the functions.

gdraynz: :ok_hand:

PR #212, opened 2017-03-28
Implement introspective lookup of template config to find data keys.

thavel:


nyuki/workflow/api/vars.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+@resource('/workflow/vars/{tid}', versions=['v1'])
+class ApiVars(DataInspector):
+
+    async def get(self, request, tid):
+        try:
+            keys = await self.required_keys(tid)
+            return Response(body=keys)
+        except TemplateError:
```
gdraynz: You can use `raise HTTPBreak` from the above method to avoid copy/paste of exception handling.
```python
from nyuki.api import HTTPBreak
raise HTTPBreak(404, {'error': 'template not found'})
```

PR #214, opened 2017-04-05
Add a frame sampler 'trace' config is true

gdraynz:


nyuki/nyuki.py
```diff
@@ -63,6 +65,11 @@ def __init__(self, **kwargs):
 
         # Initialize logging
         logging.config.dictConfig(DEFAULT_LOGGING)
+        # Debugging setup
+        if os.environ.get('DEBUG') in ('1', 'true'):
```
thavel: Can we add a key in the config file instead?

gdraynz: Yep, you can now set it as `"trace": true|false"

nyuki/bus/mqtt.py
```diff
@@ -185,9 +186,13 @@ def _regex_topic(self, topic):
         await self.client.subscribe([(topic, QOS_1)])
 
         try:
-            self._subscriptions[topic].add(callback)
+            self._subscriptions[topic].add(
+                (self._regex_topic(topic), callback),
```
thavel: You didn't update `_regex_topic()` to return a compiled regex (with `re.compile()`). At the moment, the method returns a raw string.

PR #215, opened 2017-04-10
More clever use of tukio's broker

gdraynz:


nyuki/workflow/api/workflows.py
```diff
@@ -122,24 +122,19 @@ class _WorkflowResource:
 
     def register_async_handler(self, async_topic, wflow):
         broker = get_broker()
+        topic = '{}/{}'.format(EXEC_TOPIC, wflow.uid)
```
lhavel: `'/'.join([EXEC_TOPIC, wflow.uid])` would be slightly faster; but that's ok here. I think that does not affect the overall performance in that case.

nyuki/bus/mqtt.py
```diff
@@ -213,7 +213,7 @@ def _regex_topic(self, topic):
             log.debug("MQTT unsubscription from %s->%s", topic, callback.__name__)
             self._subscriptions[topic].callbacks.remove(callback)
 
-        if callback is None or not self._subscriptions[topic]:
+        if callback is None or not self._subscriptions[topic].callbacks:
```
gdraynz: It is that way to make sure the last unsubscribe (that empties the callbacks list) remove the subscription dict entry, otherwise you need a second unsubscribe to remove the entry.

PR #216, opened 2017-04-13
WFLOW-134-lightened-history-data

RudolfFrederiksen:


nyuki/workflow/api/workflows.py
```diff
@@ -37,6 +37,36 @@ class InstanceCollection:
 
     REQUESTER_REGEX = re.compile(r'^nyuki://.*')
 
+    # TODO: Many fields may not be necessary for the frontend.
```
gdraynz: Mongodb forbid to use inclusions __and__ exclusions in a request filter, so we have to explicitely state the fields we want to retrieve.

nyuki/workflow/api/workflows.py
```diff
@@ -47,11 +77,51 @@ def __init__(self, instances_collection):
...
-    async def get_one(self, exec_id):
+    async def get_one(self, exec_id, full=False):
         """
         Return the instance with `exec_id` from workflow history.
         """
-        return await self._instances.find_one({'exec.id': exec_id}, {'_id': 0})
+        filters = self.HISTORY_FILTERS if full is False else {'_id': 0}
+        return await self._instances.find_one({'exec.id': exec_id}, filters)
+
+    async def get_task(self, task_id, full=False):
```
lhavel: As far as I understood, the `full` option won't be used by the frontend.
We'd better implement what's required. No more. No less.

PR #217, opened 2017-04-18
Split wflow instances into workflows/tasks collections

gdraynz:


nyuki/workflow/api/workflows.py
```diff
@@ -36,92 +36,89 @@ def keys(cls):
...
     }
 
-    def __init__(self, instances_collection):
-        self._instances = instances_collection
-        asyncio.ensure_future(self._instances.create_index('exec.id', unique=True))
-        asyncio.ensure_future(self._instances.create_index('exec.state'))
-        asyncio.ensure_future(self._instances.create_index('exec.requester'))
+    def __init__(self, db):
+        self._workflows = db['workflow-instances']
+        self._tasks = db['task-instances']
```
gdraynz: These lines implies a change from one `instances` collection to two `workflow-instances` and `task-instances` collections, preventing us from fetching previous history after this is merged. This should not be an issue right now.

nyuki/workflow/api/workflows.py
```diff
@@ -36,92 +36,89 @@ def keys(cls):
 class InstanceCollection:
```
lhavel: Nothing to do with this PR but that's weird to have pieces of the data model inside the web API module whereas there's a separate dedicated module for that.

nyuki/workflow/api/workflows.py
```diff
@@ -36,92 +36,89 @@ def keys(cls):
...
-        )
-        try:
-            result = result['tasks'][0]
-        except (KeyError, IndexError, TypeError):
-            return None
+        task = await self._tasks.find_one({'exec.id': task_id}, {'_id': 0})
 
         # Remove inputs and outputs if not needed.
-        if full is False:
+        if task and full is False:
```
lhavel: Why don't you leave the filtering of the results to the Mongo engine (just what you did in `_get_wflow_tasks`)?
You could add `{'inputs': False, 'outputs': False}` to the projection argument of `find_one()` and get the filtered result at once. I believe this would be less memory consuming.

nyuki/workflow/api/workflows.py
```diff
@@ -36,92 +36,89 @@ def keys(cls):
...
-        return {
-            'inputs': result['exec']['inputs'],
-            'outputs': result['exec']['outputs'],
-        }
+        task = await self._tasks.find_one({'exec.id': task_id}, {'_id': 0})
+        if task:
+            return {
+                'inputs': task['exec']['inputs'],
+                'outputs': task['exec']['outputs'],
+            }
```
lhavel: Shouldn't you add a `return None` at last (explicit better than implicit paradigm...)?

nyuki/workflow/api/workflows.py
```diff
@@ -36,92 +36,89 @@ def keys(cls):
...
+        'exec.reporting.status': 1,
+        'exec.reporting.errors': 1,
     }
 
-    def __init__(self, instances_collection):
-        self._instances = instances_collection
-        asyncio.ensure_future(self._instances.create_index('exec.id', unique=True))
-        asyncio.ensure_future(self._instances.create_index('exec.state'))
-        asyncio.ensure_future(self._instances.create_index('exec.requester'))
+    def __init__(self, db):
```
lhavel: That's weird to keep a single `InstanceCollection` class whereas there are 2 collections behind.
Why don't you split it properly into two clean classes where you could call methods from `TaskInstanceCollection` from `WorfklowInstanceCollection`

gdraynz: I can rename `InstanceCollection` to something else, but it's easier to read and use when the two collections are in the same class if they are that much linked between them.

lhavel: No need to rename it then. That breaks your own naming rules.
There's a mix between a per-collection approach (pure CRUD-based) and a true storage API (a single class to handle instance docs).
We'll have to improve that soon. Just realign the naming of your class with the others and let's merge!

PR #218, opened 2017-04-19
Separate mqtt regex topics and simple topics

gdraynz:


gdraynz:
Right, this is a "simple" bug fix.

PR #219, opened 2017-04-19
Allow dispatch of workflow async event to specific types

gdraynz:


nyuki/workflow/api/workflows.py
```diff
@@ -199,13 +199,19 @@ class _WorkflowResource:
...
         topic = '/'.join((EXEC_TOPIC, wflow.uid))
 
+        if events:
+            events = events.split(',')
+
         async def exec_handler(event):
-            # Publish the event's data
-            await self.nyuki.bus.publish(event.data, async_topic)
+            # Publish the event's data if requested.
+            if events and event.data['type'] in events:
```
lhavel: Both conditions can be merged I think:
```
if events is None or event.data['type'] in events:
    await self.nyuki.bus.publish(event.data, async_topic)
```

PR #220, opened 2017-04-26
Wait for a workflow to start before releasing the start request

gdraynz:


nyuki/workflow/api/workflows.py
```diff
@@ -317,7 +317,15 @@ class ApiWorkflows(_WorkflowResource):
         if async_topic is not None:
             self.register_async_handler(async_topic, async_events, wflow)
 
-        return Response(wfinst.report())
+        try:
```
lhavel: Well, you added this code to avoid HTTP request timeouts in case the event loop would be stuck between the `__init__()` and the call to `start()`.
If such a situation occurs, the event loop won't be able to ensure the response is sent back in a timely fashion either.
I think your code does not save us from request timeout.

lhavel: Did you run some tests to check this assertion?

lhavel: At last, we cannot ensure the response will be delivered before timeout, but the response is sent only when the workflow is started. Which is a better behavior than before.

PR #221, opened 2017-04-27
Format datetimes to UTC during serialization only

gdraynz:


nyuki/bus/reporting.py
```diff
@@ -148,10 +148,9 @@ def send_report(self, rtype, data):
             ),
             'type': rtype,
             'author': self._name,
-            'datetime': datetime.utcnow().isoformat(),
+            'datetime': datetime.utcnow(),
             'data': data
         }
-        self.check_report(report)
```
gdraynz: That was useless, no need to validate something we just built.

nyuki/utils/serialize.py
```diff
@@ -15,4 +24,4 @@ def _serialize_datetime(dt):
     """
     Datetime serializer.
     """
-    return dt.isoformat()
+    return dt.replace(tzinfo=SimpleUTC()).isoformat()
```
gdraynz: This is the main change of this issue.
Every datetime object go through this method during serialization.

nyuki/bus/reporting.py
```diff
@@ -148,10 +148,9 @@ def send_report(self, rtype, data):
             ),
             'type': rtype,
             'author': self._name,
-            'datetime': datetime.utcnow().isoformat(),
+            'datetime': datetime.utcnow(),
```
lhavel: I'd make it even simpler and safer by building a true localized datetime object this way:
```
'datetime': datetime.now(pytz.utc)
```
You'll get something like `datetime.datetime(2017, 5, 3, 12, 32, 3, 989396, tzinfo=<UTC>)` which provides the right string: `dt.isoformat()` returns `"2017-05-03T12:32:03.989396+00:00"`.

nyuki/utils/serialize.py
```diff
@@ -1,5 +1,14 @@
-from datetime import datetime
 from functools import singledispatch
+from datetime import datetime, tzinfo, timedelta
+
+
+class SimpleUTC(tzinfo):
```
lhavel: See my first comment to avoid subclassing `tzinfo`.

nyuki/utils/serialize.py
lhavel: `dt.isoformat()` is enough with my first comment

nyuki/bus/reporting.py
lhavel:
```
from datetime import datetime, timezone
datetime.now(timezone.utc)
```
does the job as well and avoid using `pytz`

nyuki/workflow/workflow.py
```diff
@@ -273,7 +271,7 @@ def new_workflow(self, template, instance, **kwargs):
             'type': event.data['type'],
             'data': event.data.get('content') or {},
             'source': source,
-            'timestamp': datetime.utcnow().isoformat()
+            'timestamp': utcnow,
```
lhavel: `utcnow()`

PR #224, opened 2017-05-16
Split bus persistence into 2 modules, do less mongo requests on publish

gdraynz:


nyuki/bus/mqtt.py
```diff
@@ -45,11 +45,12 @@ class MqttBus(Service):
                     'port': {'type': 'integer'},
                     'persistence': {
                         'type': 'object',
+                        'required': ['backend'],
                         'properties': {
-                            'backend': {
-                                'type': 'string',
-                                'minLength': 1
-                            }
+                            'backend': {'type': 'string', 'enum': [
```
lhavel: if the user follows this config scheme it does not work properly with mongo.
Other config keys are expected to make it work (`"ttl"`, `"host"`, `"ssl"`...). This must be documented here as well!

gdraynz: The thing is we only pass arguments to [the underlaying library](http://api.mongodb.com/python/current/api/pymongo/mongo_client.html#pymongo.mongo_client.MongoClient). I'll add a comment/readme.

PR #226, opened 2017-06-06
Split websocket topics to multiple subs

gdraynz:


nyuki/workflow/workflow.py
```diff
@@ -242,20 +248,47 @@ def new_workflow(self, template, instance, **kwargs):
...
+                'exec_id': source['task_exec_id'],
+                'template_id': source['task_template_id'],
+            }
+
+            if event.data['type'] == TaskExecState.progress.value:
+                # Only append full data if this is a 'task-progress'
+                payload['data'] = event.data.get('content') or {}
+                topic = '{}/reporting'.format(topic)
+
+            elif event.data['type'] == TaskExecState.end.value:
```
lhavel: Here is where I'd go for another option to filter the data to be published to the MQTT topic.
1. Get the class of the task (from `name = wflow['tasks']['task_template_id']['name']` then klass = `self.AVAILABLE_TASKS[name]`)
1. Then either add a handful method to get the filtered list of keys to publish from `data['content']` or a simple static at the class level (e.g. `WS_FILTER = None` mean "no filter, send all"; `WS_FILTER = ['quorum']` means send `'quorum'` only)
1. At last built the dict of data to send (`filtered_dict = {k:v for (k,v) in d.items() if k in klass.WS_FILTER}`)

PR #228, opened 2017-06-16
Add separated contact data from reporting in the call task

gdraynz:


nyuki/workflow/workflow.py
```diff
@@ -278,11 +279,19 @@ def new_workflow(self, template, instance, **kwargs):
...
             }
 
+            # Append full data if this is a 'task-progress'
             if event.data['type'] == TaskExecState.progress.value:
-                # Only append full data if this is a 'task-progress'
                 payload['data'] = event.data.get('content') or {}
                 topic = '{}/reporting'.format(topic)
 
+            # Custom event type for single contact updates
+            elif event.data['type'] == CONTACT_PROGRESS:
```
lhavel: That's ok to use a custom type to sort out events from notification tasks.
That way you can publish it into an extended topic.
However, I wouldn't nest contact's details into the workflow/task tree.
Contacts are not part of the exec tree, I'd expect to find contact details into the ['data'] field together with relevant progress info.
```
"data": {
    "contact": {"id": ...},
    ...
}
```
I'd also publish that kind of event into `{}/reporting/contacts/{}`.
Bear in mind tasks may process other kind of data and the client may want to split its subscriptions.

nyuki/workflow/tasks/utils/__init__.py
```diff
@@ -1,2 +1,5 @@
 from .placeholder_mapper import placeholder_mapper
 from .selectors import generate_factory_schema
+
+CONTACT_KEY = '__contact__'
```
gdraynz: Not so much a hidden key but a temporary one, only used within the code to pass useful informations. Though if we put the contact UID in the data we can indeed remove this key.

PR #229, opened 2017-06-21
Improved /instances route, improve mqtt subscription

gdraynz:


nyuki/bus/mqtt.py
```diff
@@ -206,14 +205,19 @@ def _regex_topic(self, topic):
...
                 self._subscriptions[topic].add(callback)
             except KeyError:
                 self._subscriptions[topic] = {callback}
+                sub = True
 
-        log.info('Subscribed to %s', topic)
+        # Send the subscription packet only if we were not subscribed yet
+        if sub is True:
+            await self.client.subscribe([(topic, QOS_1)])
+            log.info('Subscribed to %s', topic)
```
gdraynz: This avoid sending a subscription packet when we're already subscribed to this topic.

nyuki/bus/mqtt.py
```diff
@@ -355,11 +359,11 @@ def _regex_topic(self, topic):
                 if mqttregex.regex.match(topic):
                     log.debug("Event from topic '%s': %s", topic, data)
                     for callback in mqttregex.callbacks:
-                        asyncio.ensure_future(callback(topic, copy(data)))
+                        asyncio.ensure_future(callback(topic, data.copy()))
```
lhavel: `await` instead of `ensure_future` here also?

gdraynz: Calling subscription callbacks in tasks avoid having any interference between each callback, also this is the MQTT listening loop, so we should not slow it down by any means.

PR #237, opened 2017-08-22
Migrations of task configs using an int scheme

gdraynz:


nyuki/workflow/tasks/trigger_workflow.py
```diff
@@ -16,6 +16,19 @@
 log = logging.getLogger(__name__)
 
 
+def _migrate_1(task):
+    config = task['config']
+    task['config'] = {
+        'template': {
+            'service': 'twilio' if 'twilio' in config['nyuki_api'] else 'pipeline',
```
lhavel: Why do you set a `service` key for each task? There's already a `service` key at the workflow template level...

nyuki/workflow/tasks/trigger_workflow.py
```diff
@@ -16,6 +16,19 @@
 log = logging.getLogger(__name__)
 
 
+def _migrate_1(task):
```
lhavel: I don't feel comfortable with migrations described in referenced in a class that implements the business logic.
We'd rather design a clean db layer that deals with that stuff.
This is the major caveat I can see in your PR.

nyuki/workflow/tasks/trigger_workflow.py
```diff
@@ -16,6 +16,19 @@
...
+    config = task['config']
+    task['config'] = {
+        'template': {
+            'service': 'twilio' if 'twilio' in config['nyuki_api'] else 'pipeline',
+            'id': config['template'],
+            'draft': config.get('draft', False),
```
lhavel: Same question with `draft`

nyuki/workflow/tasks/trigger_workflow.py
gdraynz: It points to the template to be triggered by the `trigger_workflow` task.

nyuki/workflow/api/templates.py
```diff
@@ -73,19 +102,27 @@ def __init__(self, templates_collection, metadata_collection):
                 template.update(metadatas[template['id']])
 
         if latest is False and draft is False:
+            if full is True:
+                for template in templates:
+                    await self._migrate(template)
             return templates
 
         # Retrieve the latest versions + drafts
         lasts = {}
```
gdraynz: The frontend does not use it, I sometimes do when in development.

nyuki/workflow/api/templates.py
```diff
@@ -41,6 +43,33 @@ def __init__(self, templates_collection, metadata_collection):
...
+                # Do one step of migration.
+                do_migration(task)
+                task['scheme'] = scheme
+                must_save = True
+                log.info(
+                    'Migrated task %s of template %s from %s to %s',
+                    task['id'], template['id'], old_scheme, scheme,
+                )
+
+        if must_save:
```
gdraynz: This is to ensure we only replace the database document if at least one task was migrated.

nyuki/workflow/api/templates.py
lhavel: I'm ok to improve Mongo requests in a further issue.
But we should not leave unused options in the code. If you need full details of a workflow template the `get` method is the one you should use.

nyuki/workflow/api/templates.py
lhavel: yep, and the `if migrations[scheme:]` condition evaluates to true only if there's at least one migration to apply and no exception occurred during the migration process. This is what you want isn't it?

nyuki/workflow/tasks/trigger_workflow.py
gdraynz: This is from my point of view the safest and simplest spot to never miss a migration when one of our tasks has to change: everything in the corresponding task's module.
Please note that the actual "database model" of each task really is the attached JsonSchema.
If I agree that our database model must be better controlled, I find this specific task migration system ok.

lhavel: The task holder class is meant to execute the business logic.
It is neither meant to define the data served by the HTTP API and nor the data stored into the db backend.
I wish we get back to more future-proof design patterns.
The JSON schema may no longer be part of the task holder class in a near future...
I understand your _"everything in one place is simpler, safer"_ but that mixes distinct concerns

nyuki/workflow/api/templates.py
```diff
@@ -41,6 +43,33 @@ def __init__(self, templates_collection, metadata_collection):
             [('id', DESCENDING), ('draft', DESCENDING)]
         ))
 
+    async def _migrate(self, template):
+        # Check and do migrations if needed
+        must_save = False
+        for task in template['tasks']:
+            old_scheme = task.get('scheme', 0)
```
lhavel: `old_scheme` => `cur_scheme`, the scheme value may not change

nyuki/workflow/api/templates.py
lhavel: My bad. I didn't read the condition at the right indentation level.

nyuki/workflow/api/templates.py
```diff
@@ -109,6 +146,37 @@ def __init__(self, templates_collection, metadata_collection):
                 for template in templates:
                     template.update(metadatas[0])
 
+        for template in templates:
```
lhavel: I didn't get it at once but each time you read templates, you try to migrate it!
Which means you need to walk through the list of templates and for each template you walk through the list of tasks to ensure there's nothing to migrate.
This is likely to add a significant overhead and may lead to performance issues.

lhavel: We may have to _flag_ each document as _migrated_ once it's done...

lhavel:
We identified several problems in the approach proposed by @gdraynz.
This branch won't be merged.
New issues and branches shall be created to address migrations with a more future-proof approach.

PR #239, opened 2017-09-08
New layer on workflow history

gdraynz:


nyuki/workflow/db/workflow_instances.py
```diff
@@ -14,12 +14,12 @@
 
 class Ordering(Enum):
```
lhavel: unused imports `uuid4` and `DuplicateKeyError`

PR #253, opened 2018-02-13
Add workflow CSV export route

gdraynz:


nyuki/workflow/api/instances.py
```diff
@@ -365,6 +365,24 @@ class ApiWorkflowHistory:
...
+@resource('/workflow/history/{uid}/csv', versions=['v1'])
+class ApiWorkflowHistoryCSV:
+
+    async def get(self, request, uid):
+        timezone = request.GET.get('timezone', 'utc')
+        csv = await self.nyuki.storage.get_workflow_csv(uid, timezone)
+        if csv is None:
+            # If nothing to report, return 404
+            return Response(status=404)
+        return Response(
```
Ketouem: As the csv file could be potentially big, waiting for a fully generated export could trigger a timeout error on the client's side. One way to circumvent this is to stream the data as they are generated, this implies making sone modification into `get_workflow_csv` to make it return a generator on which we'll iterate and feeding https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.StreamResponse

PR #262, opened 2019-02-06
NYPI-124: Add 'schema' attribute to workflow templates

gdraynz:


nyuki/workflow/api/instances.py
```diff
@@ -144,7 +143,7 @@ class ApiWorkflows(_WorkflowResource):
 
         # Keep full instance+template in nyuki's memory
         wfinst = self.nyuki.new_workflow(
-            template, wflow,
+            wf_tmpl.as_dict(), wflow,
```
lhavel: Why don't you use the original dict `template` instead of rebuilding it from the `wf_tmpl` object? It will consume more CPU/memory.

lhavel: See my comment https://github.com/optiflows/tukio/pull/72#pullrequestreview-205908785.

gdraynz: We can indeed assume the templates will be migrated and the `schema` attribute will be present in the database.

gdraynz: The `schema or 1` is simply a default value to this attribute, I don't want to permanently add a `schema` in all my new templates.

PR #266, opened 2019-05-28
Put tenant_id in database name for new installations

gdraynz:


nyuki/workflow/db/storage.py
```diff
@@ -34,14 +35,21 @@ def __init__(self):
...
         self._validate_on_start = validate_on_start
+        self._db_name = database
+
+    async def index(self):
+        """
+        Try to connect to mongo and index all the collections.
+        """
+        if self._db_name not in await self._client.list_database_names():
+            # If the old mongo DB does not exist, use the new tenant format
+            self._db_name = f"{os.environ.get('TENANT_ID', 'dev')}#{self._db_name}"
```
lhavel: If there's no tenant ID it should raise an error (no default value). Otherwise it's too easy to forget or misconfigure the tenant ID.

PR #267, opened 2019-08-12
[NYUKI-221] Global cleanup of nyuki's code and librairies

gdraynz:


nyuki/logs.py
```diff
@@ -1,41 +1,29 @@
...
-        "websockets": {
-            "level": "WARNING"
-        }
     },
-    "disable_existing_loggers": False
+    'root': {'handlers': ['console'], 'level': 'INFO'},
+    'loggers': {
+        'asyncio': {'handlers': ['loop_stderr'], 'level': 'WARNING'},
+        'gmqtt': {'level': 'WARNING'},
+        'slixmpp': {'level': 'WARNING'},
```
lhavel: As far as I know we no longer use `slixmpp`

nyuki/config.py
```diff
@@ -47,66 +24,77 @@ def merge_configs(*configs):
...
+                'File %s does not exist, using default.json',
+                self._location,
+            )
+            self._location = self.DEFAULT_CONF_FILE
+            conf = json.loads(self._location.read_text())
+        return merge_configs(self.BASE_CONF, conf)
+
+    def _get_from_s3(self):
+        """
+        TODO: Use boto3 ?
```
lhavel: Indeed, use boto3 instead of the awscli. I found a snippet in seconds:
```python
import boto3

s3 = boto3.resource('s3')

obj = s3.Object(bucket, key)
obj.get()['Body'].read().decode('utf-8')
```

nyuki/nyuki.py
```diff
@@ -35,62 +32,30 @@ class Nyuki:
...
-        bus_config = self._config.get('bus')
-        if bus_config is not None:
-            bus_service = bus_config.get('service', 'mqtt')
-            if bus_service == 'mqtt':
-                self._services.add('bus', MqttBus(self))
-
-        self.is_stopping = False
+        self.services = ServiceManager(self)
+        self.services.add('api', Api(self))
+        self.services.add('bus', MqttBus(self))
```
lhavel: It's convenient to start a nyuki standalone for testing/dev without even a broker.
If there's no `bus` key in the config, the nyuki should not event try to connect.

nyuki/nyuki.py
```diff
@@ -100,196 +65,126 @@ def __getattribute__(self, name):
...
+        await nyuki.start()
+        await nyuki.wait()
 
+    async def start(self):
+        """
+        Start the nyuki's services.
+        """
+        self.__stop_event.clear()
+        # Set signal handlers
+        loop = asyncio.get_event_loop()
```
lhavel: use the `self.loop` property here

nyuki/nyuki.py
```diff
@@ -35,62 +32,30 @@ class Nyuki:
...
+        self._configuration = Configuration(
+            command_args.config or os.environ.get('CONF_LOCATION'),
+        )
+
+        self._log_level = command_args.logging
+        self.configure_logging(DEFAULT_LOGGING)
+        self.configure()
 
         # Setup stack sampling
         self._sampler = None
```
lhavel: It looks like this attribute `self._sampler` is not used anywhere in the code...

nyuki/nyuki.py
```diff
@@ -100,196 +65,126 @@ def __getattribute__(self, name):
...
+        # Set logging level
+        if self._log_level:
+            logging.getLogger().setLevel(self._log_level)
 
     async def setup(self):
         """
         First thing called when starting the event loop, coroutine or not.
         """
-        log.warning('Setup called, but not overridden')
+        log.debug('Setup called, but not overridden')
```
lhavel: The warning level was fine

nyuki/utils/__init__.py
```diff
@@ -1,4 +1,9 @@
+import os
+
 from .dtutils import from_isoformat, utcnow
 from .evaluate import safe_eval, ConditionBlock
-from .serialize import serialize_object
+from .serialize import serialize_object, json_default
 from .transform import Converter
+
+
+IS_DEBUG = os.environ.get('DEBUG') == '1'
```
lhavel: The log level should be all that must be defined

nyuki/services.py
```diff
@@ -5,24 +5,23 @@
 log = logging.getLogger(__name__)
 
 
-class Service(object):
+class Service:
 
     async def start(self, *args, **kwargs):
-        raise NotImplementedError
+        pass
```
lhavel: Why did you remove the `raise NotImplementedError`; are there situations where one of these methods might not be overridden?

nyuki/api/api.py
```diff
@@ -16,201 +14,78 @@
...
-        except HTTPBreak as exc:
-            return Response(exc.body, status=exc.status)
-
-        if isinstance(capa_resp, Response):
-            return capa_resp
-        return Response()
-
-    return middleware
-
+            return await handler(**self.request.match_info)
```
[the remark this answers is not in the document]
gdraynz: This allows to create and expand a custom api error handler using the `singledispatch` method, and return JSON errors more easily.

nyuki/bus/mqtt.py
```diff
@@ -62,64 +37,41 @@ def topics(self):
...
-            except asyncio.CancelledError:
-                log.debug('future cancelled: %s', future)
-        self.connect_future = asyncio.ensure_future(self._run())
-        self.connect_future.add_done_callback(cancelled)
+                await self.client.connect(
+                    self._dsn.host,
+                    self._dsn.port,
+                    (self._dsn.scheme == 'mqtts'),
+                )
+            except ConnectionRefusedError as exc:
```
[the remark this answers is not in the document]
gdraynz: I remember talks about nyukis that should never stop if a third-party service is down but simply try to reconnect but yes stopping it after a while may be better

nyuki/config.py
gdraynz: The thing is this method is not really intended to be used right now as we're kind of between two ways of configuring our nyukis and I didn't want to add the (big) boto3 dependency right now.

nyuki/nyuki.py
gdraynz: This is purely an implementation thing that should be relevant only when developing a nyuki and not during its normal execution.

nyuki/services.py
gdraynz: Same as above, this is an implementation specificity and some services (like the StackSampler here) may not need to implement these 3 methods, yet it's easier to call the 3 methods on all available services at once. I added a debug log to simply notice this.

nyuki/nyuki.py
lhavel: Then it should be fixed during the development phase and never happen in production.
The debug level just hides real problems or things we might have missed during the dev.

nyuki/utils/__init__.py
lhavel: As far as I can see it is used only to insert exception details into the JSON response in "DEBUG mode". Which is not a good practice. The logs in DEBUG level are best suited for that and does not cost more.

nyuki/api/errors.py
```diff
@@ -0,0 +1,29 @@
...
+from nyuki.utils import IS_DEBUG
+
+
+log = logging.getLogger(__name__)
+
+
+@singledispatch
+def api_error_handler(exc, request):
+    log.exception(exc)
+    if IS_DEBUG:
```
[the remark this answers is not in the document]
gdraynz: Exceptions are always logged, but will also be showed in the HTTP response if the nyuki runs in debug

nyuki/nyuki.py
gdraynz: No these are methods that are always called by the nyuki library, but not necessarily used by every nyuki. If my nyuki does not have any initial setup or teardown cleaunp to do, I don't care about this warning message popping every time.

nyuki/utils/__init__.py
gdraynz: This is also a small helper to add all sorts of things when developing a nyuki that have some debug specifics, instead of having to fetch the environment variable by hand.

nyuki/api/errors.py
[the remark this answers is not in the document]
lhavel: Please remove exception details from HTTP response this is definitely a very bad practice.
APIs should never expose implementation details.

nyuki/config.py
lhavel: Ok then remove this method: why should we keep code that is not used. It just make the codebase harder to maintain.

nyuki/nyuki.py
lhavel: ok let's keep it like that

